PR #1, opened 2024-06-27
Add Realisation Specification

lispandfound:
## Changes
- Added a comprehensive proposal for a new realisation format, consolidating multiple files (`wiki/Realisations.md`).
- Introduced `workflow/realisations.py` reading and validating new realisation files.
- Implemented a realisation schema (incomplete at the moment).
- Updated project metadata in `pyproject.toml` and added new dependencies in `requirements.txt`.

## Details
This PR begins to address the need for a unified format for realisation specifications, aiming to simplify workflow and improve reproducibility in earthquake simulations. The new wiki page `wiki/Realisations.md` outlines the structure and advantages of the proposed JSON-based realisation specification, which replaces multiple YAML and CSV files. It details the components of a realisation and emphasizes readability, writability, referential transparency, and validation. Additionally, `workflow/realisations.py` provides initial schema validation and structure for realisation reading.

## Not Yet Implemented
- `RealisationBuilder` interface for building realisations.

## Example
```python
from workflow import realisations

print(realisation.read_realisation_file('realisation.json'))
```

## Example Realisation

``` json
{
  "name": "Realisation name",
  "version": "5",
  "srf_generation_parameters": {
    "genslip_dt": 0.05,
    "genslip_seed": 1,
    "genslip_version": "5.4.2",
    "srfgen_seed": 1
  },
  "sources": {
    "Acton":   {
      "type": "point",
      "coordinates": {
        "latitude": -45.0,
        "longitude": 172.0,
        "depth": 0.0
      },
      "strike": 90.0,
      "dip": 20.0,
      "dip_dir": 0.0
    }
  },
  "domain_parameters": {
    "resolution": 0.1,
    "model_centroid": {
      "latitude": -45.0,
      "longitude": 172.0
    },
    "model_width": 100.0,
    "model_length": 100.0,
    "model_depth": 20.0
  },
  "rupture_propagation": {
    "Acton": {
      "parent": null,
      "hypocentre": {
        "s": 0.5,
        "d": 0.5
      },
      "magnitude": 7.0,
      "rake": 110.0
    }
  }
}
```


lispandfound:
As an example of the schema errors in action. Consider the following specification that is only slightly different from the example in the PR
```json

```
When you read this realisation with `read_realisation_file` you get the following error
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/var/home/jake/src/workflow/workflow/realisations.py", line 412, in read_realisation_file
    return REALISATION_SCHEMA.validate(json.load(realisation_file_handle))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/jake/src/source_modelling/.venv/lib/python3.12/site-packages/schema.py", line 544, in validate
    raise SchemaError(
schema.SchemaError: Key 'srf_generation_parameters' error:
Key 'genslip_version' error:
Or('5.4.2') did not validate '5.4.1'
'5.4.2' does not match '5.4.1'
```
This makes it clear that:
1. The realisation specification is incorrect (rather than the JSON syntax being wrong)
2. The error is with the `srf_generation_parameters.genslip_version` key
3. The error is that the genslip version can only be 5.4.2

workflow/realisations.py
```diff
@@ -0,0 +1,412 @@
...
+# schema.SchemaError: <lambda>(-12) should evaluate to True
+#
+# But using the function `is_positive` we instead have
+#
+# And(float, is_positive).validate(-12)
+# schema.SchemaError: is_positive(-12) should evaluate to True
+#
+# So using short functions with names improves the error reporting
+# from the library.
+#
```
sungeunbae, line 85: brilliant

sungeunbae (approved):

workflow/realisations.py
```diff
@@ -0,0 +1,412 @@
...
+        Literal("model_length", description="The length of the model (in km)"): And(
+            float, is_positive
+        ),
+        Literal("model_depth", description="The depth of the model (in km)"): And(
+            float, is_positive
+        ),
+    }
+)
+
+REALISATION_SCHEMA = Schema(
```
claudio525, line 369: This schema is specific for type 5 only? Might be good to reflect this in the name?

claudio525 (commented):
Nice work!

workflow/realisations.py
lispandfound: It's type-5 for now. The idea is to eventually port it to all types (hence the name). There is nothing specific to type-5 about the schema.

lispandfound: The version of the realisation format is "5", but it could easily be 1.

lispandfound:
I have a had a slightly better idea on how to structure this code that I think will make it nicer to use. Will update this later this afternoon when it's good to go.

lispandfound:
This pull is ready to review again. I've updated the structure of the module to reflect the fact that realisations not built all at once. Accordingly, there is no longer one big `Realisation` object. Instead I have created smaller dataclasses for each section, `DomainParameters`, `SRFConfig`, etc. You can then read these separately from the realisation file. Writing to realisation files is now also modular. The whole system is now designed to be extended as this format consumes more of the workflow. See the  `Realisations.md` wiki page for a detailed overview of how it works. 

For viewing wiki diffs, I just learned you can do rich-text output. Click the not at all obvious button highlighted in the screenshot below and the markdown will format like the wiki is. This will stop me creating the PDFs alongside every PR I make.
image

lispandfound:
Here is an updated example of a realisation file

```json
{
  "metadata": {
    "name": "Test Realisation",
    "version": "1"
  },
  "domain": {
    "resolution": 0.0,
    "centroid": {
      "latitude": -43.53092,
      "longitude": 172.63701
    },
    "width": 100.0,
    "length": 100.0,
    "depth": 40.0
  },
  "srf": {
    "genslip_dt": 1.0,
    "genslip_seed": 1,
    "genslip_version": "5.4.2",
    "srfgen_seed": 1
  },
  "sources": {
    "source_geometries": {
      "Acton": {
        "type": "point",
        "coordinates": {
          "latitude": -43.53092,
          "longitude": 172.63701,
          "depth": 0.0
        },
        "length": 100.0,
        "strike": 90.0,
        "dip": 60.0,
        "dip_dir": 180.0
      },
      "Nevis": {
        "type": "plane",
        "corners": [
          { "longitude": 172.52014, "latitude": -43.42214, "depth": 0 },
          { "longitude": 172.52014, "latitude": -43.5768, "depth": 0 },
          { "longitude": 172.68364, "latitude": -43.42214, "depth": 1000 },
          { "longitude": 172.68364, "latitude": -43.5768, "depth": 1000 }
        ]
      },
      "Nevis2": {
        "type": "fault",
        "corners": [
          { "longitude": 172.52014, "latitude": -43.42214, "depth": 0 },
          { "longitude": 172.52014, "latitude": -43.5768, "depth": 0 },
          { "longitude": 172.68364, "latitude": -43.42214, "depth": 1000 },
          { "longitude": 172.68364, "latitude": -43.5768, "depth": 1000 },

          { "longitude": 172.52014, "latitude": -43.42214, "depth": 0 },
          { "longitude": 172.52014, "latitude": -43.5768, "depth": 0 },
          { "longitude": 172.68364, "latitude": -43.42214, "depth": 1000 },
          { "longitude": 172.68364, "latitude": -43.5768, "depth": 1000 }
        ]
      }
    }
  }
}
```

workflow/realisations.py
```diff
@@ -268,145 +440,237 @@ def corners_to_array(corners_spec: list[dict[str, float]]) -> np.ndarray:
...
-        ): "point",
-        Literal(
-            "coordinates", description="The coordinates of the point source"
-        ): LAT_LON_DEPTH_SCHEMA,
-        Literal("strike", description="The strike bearing of the point source"): And(
-            float, is_valid_bearing
-        ),
-        Literal("dip", description="The dip angle of the point source"): And(
-            float, is_valid_bearing
+    And(
```
joelridden, line 444: Would it make sense to have these SCHEMA defenitions in a separate file just so this realisations.py doesn't blow up to be massive when extended?

workflow/realisations.py
```diff
@@ -0,0 +1,676 @@
...
+    RupturePropagationConfig,
+    RealisationMetadata,
+]
+
+
+class RealisationParseError(Exception):
+    pass
+
+
+def read_config_from_realisation(config: type, realisation_ffp: Path) -> LoadableConfig:
```
joelridden, line 614: Is there any way to load all of the different parameters at once without needing to call this multiple times for each of the different subsections?

joelridden (commented):
Just some thoughts to maybe discuss, code looks great though. Will for sure improve workflow

workflow/realisations.py
sungeunbae: Good point @joelridden, makes me curious too!

workflow/realisations.py
lispandfound: yeah maybe a schemas module?

workflow/realisations.py
lispandfound: So the short answer is no. I have thought about this a bit. The truth is that I don't think there is a massive need for it. No part of the workflow needs to know about the SRF generation parameters and the HF parameters, for example. However, there is some scope to extend this by, say, making `config` optionally a list and then returning a list of `LoadableConfig` in that case. I'll write the code for that now.

claudio525 (approved):
Good idea to break it into multiple pieces! Nice work!

sungeunbae (approved):
Excellent.

workflow/realisations.py
```diff
@@ -0,0 +1,382 @@
+"""The realisations module defines the schema for the realisation file format.
+
+The configuration schemas are contained in `_REALISATION_SCHEMAS`. The
+schemas loosely validates the input data. Input bounds checking is
```
AndrewRidden-Harper, line 4: Possible typo: The schemas loosely (validates --> validate)  the input data.

workflow/realisations.py
```diff
@@ -0,0 +1,382 @@
...
+from source_modelling.rupture_propagation import JumpPair
+from source_modelling.sources import IsSource
+from workflow import schemas
+
+
+def to_name_coordinate_dictionary(
+    coordinate_array: np.ndarray,
+    coordinate_names: list[str] = ["latitude", "longitude", "depth"],
+) -> Union[dict[str, float], list[dict[str, float]]]:
+    """Convert an array of coordinates values into a (list of) dictionaries tagged with coordinate names.
```
AndrewRidden-Harper, line 48: Should this be coordinates --> coordinate ?

workflow/schemas.py
```diff
@@ -0,0 +1,352 @@
...
+
+    Parameters
+    ----------
+    corners : np.ndarray
+        The corners to validate.
+
+    Returns
+    -------
+    bool
+        Returns true if the last column of the corners is
```
AndrewRidden-Harper, line 114: Very minor but this should probably be True with a T

AndrewRidden-Harper (approved):
Great work! Very well thought out code and documentation

joelridden (dismissed):

claudio525 (approved):

PR #5, opened 2024-10-15
Scientific Defaults

lispandfound:
This PR introduces the new workflow changes for scientific defaults. These defaults are an accumulation of all of the magic numbers I could find in the old workflow codebase. These defaults and read and written to realisations by the workflow stages.

## Why These Changes?

The old workflow kept defaults across a number of different files. Some default values, such as those for the high frequency simulations, were recorded in code and not recognised as values with scientific significance for simulations. Consolidating the defaults has three advantages:

1. Simulations are more easily reproducible because the state can no longer be inconsistent between the defaults and the code. That is, it is no longer possible for the codebase to be using scientific defaults from 2022 for velocity modelling and EMOD3D but then using the latest HF code from last week (which has HF defaults in the old workflow).
2. Simulation defaults are easily shared and tweaked because they live in just one place.
3. It is easier to swap out defaults and run 200m instead of 400m for the same realisation.

## Summary of the files involved

The `workflow/default_parameters` directory contains four sets of defaults, `24.2.2.1`, `24.2.2.2`, `24.2.2.4` and `develop` respectively. The first three represent a closest attempt to copy over the defaults from the old workflow. They should not deviate significantly from the old workflow default values. The `develop` defaults are defaults for a 400m simulation designed for testing. Develop defaults are not designed to be used for any real experiments.

The `defaults.py` module contains code to load the defaults from a version (valid versions given in `DefaultsVersion`). The defaults are loaded by the updated realisations module (not part of this PR).


lispandfound:
I will introduce test cases for the defaults version module when I add the changes to the realisation configuration

workflow/defaults.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns
+    -------
+    dict
+        A dictionary containing the default parameters loaded from the YAML file.
+        The keys are strings representing parameter names, and the values can be
+        integers, floats, or strings depending on the parameter.
+    """
+    if version == DefaultsVersion.develop:
+        defaults_package = importlib.import_module(
```
joelridden, line 35: Do these need to be packages? Could these just be files in the default_parameters directory and then have a mapping dictionary for the StrEnum to the filename and just load that? Would make this simpler imo

workflow/default_parameters/v24_2_2_4/defaults.yaml
```diff
@@ -0,0 +1,268 @@
...
+  version: "2.06"
+  topo_type: "SQUASHED_TAPERED"
+  vs30: 500.0
+  s_wave_velocity: 3500.0
+  pgv_interpolants: [[3.5, 0.015], [4.1, 0.0375], [4.7, 0.075], [5.2, 0.15], [5.5, 0.25], [5.8, 0.4], [6.2, 0.7], [6.5, 1.0], [6.8, 1.35], [7.0, 1.65], [7.4, 2.1], [7.7, 2.5], [8.0, 3.0]]
+bb:
+  flo: 0.5
+  dt: 0.02
+  fmidbot: 0.5
+  fmin: 0.25
```
joelridden, line 126: These seem different to what I saw in root_defaults.yaml 20.4.1.4 fmin: 0.2 and fmidbot: 0.25, as well as dt: 0.005. Do you know why these would be different? Not gonna compare everything but these stuck out to me

joelridden (commented):
Few comments. I wonder if we had 1 default config and then 24_2_2_1 for example would just replace values that matter for 100m? Could reduce the amount of config line duplicates, not sure if this limits us at all though

workflow/default_parameters/v24_2_2_4/defaults.yaml
lispandfound: That `dt` value is incorrect for the old root_defaults. The value `dt / 20` is what is used for EMOD3D to ensure stability, that value seems to be the `dt` value used for 100m simulations instead of 400m simulations. As for `fmin` and `fmidbot` I have no idea, but I would guess that maybe the root defaults is not a good place to look if it may have been copied from the 100m defaults.

workflow/defaults.py
lispandfound: A simpler approach again would perhaps be to simply declare the defaults as variables in the `defaults.py` module. There is a trade-off though: if they're in the code it reduces the readability  somewhat for non-programmers.

Aside from that importlib approach is recommended way (per the stdlib docs) to load external resources. If we do decide to stick with external files for the defaults I would prefer to use importlib rather than `__file__` or similar.

lispandfound:
Part of the problem regarding the hierarchical approach that Joel suggests for YAML defaults is that I am not sure what variables are dependent on the resolution and what aren't. So it's hard to know what to override. Secondly, I am a fan of keeping things quite simple for defaults so that you can just pass one file around, and publish just one file in papers to show what the settings for simulations were.

claudio525 (approved):

workflow/default_parameters/v24_2_2_4/defaults.yaml
lispandfound: There are a few hints in the codebase that `dx/20` is implied: 

1. It is consistent with the values for 200m and 100m that I copied from the root defaults of the old simulation workflow
2. Places like https://github.com/ucgmsim/Pre-processing/blob/a512a1866f3edd6425a2b68d9cf96ce5a39f351c/SrfGen/slurm/gcmt/GCMT2SRF_VM.sl#L43 where we generate SRFs for 400m simulations with `dt = 0.4 / 20 = 0.02`

sungeunbae (approved):

AndrewRidden-Harper (approved):

PR #6, opened 2024-10-16
Updated Realisations

lispandfound:
# Realisations

The following PR updates the realisations for the new workflow. The API for realisations has changed since the last pull request for realisations. The new workflow realisations aim to be a well-documented, modular, and self-contained method for simulation configuration. The idea is that the scientific defaults version, container version, and realisation file should completely determine the simulation output up to hardware-level floating point mathematics differences.

> [!NOTE]
> Only the `realisations.py`, `test_realisation.py`, and `schemas.py` files need to be reviewed. The defaults are otherwise as they are in #5 , and are there to allow the tests to pass.

## Using the Realisations Module

Reading and writing realisations involves using one of the classes to read and write realisation configurations. To retrieve the domain parameters of the realisation for example you would write something like

``` python
# For a realisation with contents
#  {
#    "domain": {
#        "resolution": 0.2,
#        "domain": [
#                  { "latitude": -44.354733203836375, "longitude": 168.32522375230945 },
#                  { "latitude": -45.860251409924196, "longitude": 170.71966358185924 },
#                  { "latitude": -42.084729432307505, "longitude": 174.9917352035785 },
#                  { "latitude": -40.66798264759381, "longitude": 172.63541186873076 }
#        ],
#        "depth": 50.0,
#        "duration": 60.0,
#        "dt": 0.01
#    }
# }
#
#
#
>>> from workflow.realisations import DomainParameters
>>> domain_parameters = DomainParameters.read_from_realisation('realisation.json')
DomainParameters(
        resolution=0.2,
        domain=BoundingBox(...)
        depth=50.0,
        duration=60.0,
        dt=0.01
)
>>> domain_parameters.write_to_realisation('realisation.json')
```

Realisations can load default values for some configuration sections using `read_from_realisation_or_defaults`

``` python
# For a realisation with contents
#  {
#    "domain": {
#        "resolution": 0.2,
#        "domain": [
#                  { "latitude": -44.354733203836375, "longitude": 168.32522375230945 },
#                  { "latitude": -45.860251409924196, "longitude": 170.71966358185924 },
#                  { "latitude": -42.084729432307505, "longitude": 174.9917352035785 },
#                  { "latitude": -40.66798264759381, "longitude": 172.63541186873076 }
#        ],
#        "depth": 50.0,
#        "duration": 60.0,
#        "dt": 0.01
#    }
# }
#
# NOTE: There is no velocity model parameters in this realisation
#
>>> from workflow.realisations import VelocityModelParameters
>>> from workflow.defaults import DefaultsVersion
>>> VelocityModelParameters.read_from_realisation_or_defaults('realisation.json', DefaultsVersion.v24_2_2_2)
VelocityModelParameters(
    min_vs=0.5,
    version='2.06',
    topo_type='SQUASHED_TAPERED',
    dt=0.01,
    ds_multiplier=1.2,
    resolution=0.2,
    vs30=500.0,
    s_wave_velocity=3500.0,
    pgv_interpolants=array([
      [3.5, 0.015],
      [4.1, 0.0375],
      [4.7, 0.075],
      [5.2, 0.15],
      [5.5, 0.25],
      [5.8, 0.4],
      [6.2, 0.7],
      [6.5, 1.0],
      [6.8, 1.35],
      [7.0, 1.65],
      [7.4, 2.1],
      [7.7, 2.5],
      [8.0, 3.0]
    ])
)
# These values were read from the 24.2.2.2 defaults yaml file.
```

Realisations have rudimentary value checking applied in the `schemas` module, which also defines the schema for each realisation section. For instance, here is the schema for the velocity model parameters.

``` python
VELOCITY_MODEL_SCHEMA = Schema(
    {
        Literal(
            "min_vs",
            description="The minimum velocity (km/s) produced in the velocity model.",
        ): And(float, is_positive),
        Literal("version", "Velocity model version"): "2.06",
        Literal("topo_type", "Velocity model topology type"): str,
        Literal("dt", "Velocity model timestep resolution"): And(float, is_positive),
        Literal("ds_multiplier", "Velocity model ds multiplier"): And(
            float, is_positive
        ),
        Literal("resolution", "Velocity model spatial resolution"): And(
            float, is_positive
        ),
        Literal("vs30", "VS30 value"): And(float, is_positive),
        Literal("s_wave_velocity", "S-wave velocity"): And(float, is_positive),
        Literal('pgv_interpolants', 'PGV interpolants to estimate domain size'): And([[And(float, is_positive)]], Use(np.array))
    }
)
```

Adding a new realisation configuration is done by inheriting from the `RealisationConfiguration` class and then defining the key in the json file, the schema to validate against, and the values to store. Overriding the `to_dict` method lets you change the JSON serialisation behaviour. Here is an example of that for the velocity model parameters.

``` python
@dataclasses.dataclass
class VelocityModelParameters(RealisationConfiguration):
    """Parameters defining the velocity model."""

    _config_key: ClassVar[str] = "velocity_model"
    _schema: ClassVar[Schema] = schemas.VELOCITY_MODEL_SCHEMA

    min_vs: float
    """The minimum velocity in the velocity model."""
    version: str
    """The velocity model version."""
    topo_type: str
    """The topology type of the velocity model."""
    dt: float
    """The velocity model time resolution."""
    ds_multiplier: float
    """The ds multiplier used to adjust simulation duration."""
    resolution: float
    """The resolution of the velocity model (in kilometres)."""
    vs30: float
    """The reference vs30 value for duration estimation."""
    s_wave_velocity: float
    """The s-wave velocity."""
    pgv_interpolants: npt.NDArray[np.float32]
    """PGV interpolation nodes between rupture magnitude and PGV target values."""

    def to_dict(self) -> dict:
        """
        Convert the object to a dictionary representation.

        Returns
        -------
        dict
            Dictionary representation of the object.
        """
        _dict = dataclasses.asdict(self)
        _dict["pgv_interpolants"] = _dict["pgv_interpolants"].tolist()
        return _dict
```

Some values for the realisation configuration are computed simply using values stored in the file. These values are now computed using Python properties rather than being stored in the realisation. For example the `nx`, `ny`, and `nz` values for the domain parameters are properties computed from the domain.

``` python
@dataclasses.dataclass
class DomainParameters(RealisationConfiguration):
    """Parameters defining the spatial and temporal domain for simulation."""

    _config_key: ClassVar[str] = "domain"
    _schema: ClassVar[Schema] = schemas.DOMAIN_SCHEMA

    resolution: float
    """The simulation resoultion in kilometres."""
    domain: BoundingBox
    """The bounding box for the domain."""
    depth: float
    """The depth of the domain (in metres)."""
    duration: float
    """The simulation duration (in seconds)."""
    dt: float
    """The resolution of the domain in time (in seconds)."""

    @property
    def nx(self) -> int:
        """int: The number of x coordinate positions in the discretised domain."""
        return int(np.round(self.domain.extent_x / self.resolution))

    @property
    def ny(self) -> int:
        """int: The number of y coordinate positions in the discretised domain."""
        return int(np.round(self.domain.extent_y / self.resolution))

    @property
    def nz(self) -> int:
        """int: The number of z coordinate positions in the discretised domain."""
        return int(np.round(self.depth / self.resolution))

    def to_dict(self) -> dict:
        """
        Convert the object to a dictionary representation.

        Returns
        -------
        dict
            Dictionary representation of the object.
        """
        param_dict = dataclasses.asdict(self)
        param_dict["domain"] = to_name_coordinate_dictionary(
            self.domain.corners,
        )
        return param_dict
```

Having these as properties means it is impossible to have the `nx`, `ny` and `nz` values be inconsistent with the domain.

## Testing

Inline with the discussions on testing earlier in the year, there is extensive testing for every line of the realisation and schema module. The tests are located in the `test_realisation.py` file. The tests ensure:

1. Every realisation section is read and written in the correct way from the JSON file.
2. Every defined defaults version is readable.

### DIFF
--- a/tests/test_realisation.py
+++ b/tests/test_realisation.py
@@ -1,0 +1,403 @@
+# NOTE: this fix contains tests that mirror the Realisations wiki. If
+# these tests fail, you should update the wiki if necesary to ensure
+# it stays consistent with the codebase.
+import json
+from pathlib import Path
+
+import numpy as np
+import pytest
+import schema
+from velocity_modelling import bounding_box
+
+from source_modelling import rupture_propagation
+from workflow import defaults, realisations
+
+
+def test_bounding_box_example(tmp_path: Path):
+    domain_parameters = realisations.DomainParameters(
+        resolution=0.1,  # a 0.1km resolution
+        domain=bounding_box.BoundingBox.from_centroid_bearing_extents(
+            centroid=np.array([-43.53092, 172.63701]),
+            bearing=45.0,
+            extent_x=100.0,
+            extent_y=100.0,
+        ),
+        depth=40.0,
+        duration=60.0,
+        dt=0.005,
+    )
+    realisation_ffp = tmp_path / "realisation.json"
+    domain_parameters.write_to_realisation(realisation_ffp)
+    with open(realisation_ffp, "r") as realisation_handle:
+        assert json.load(realisation_handle) == {
+            "domain": {
+                "resolution": 0.1,
+                "domain": [
+                    {"latitude": -43.524793866326725, "longitude": 171.76204128885567},
+                    {"latitude": -44.16756820707226, "longitude": 172.63312824122775},
+                    {"latitude": -43.53034935969409, "longitude": 173.51210368762364},
+                    {"latitude": -42.894200350955856, "longitude": 172.64076673694242},
+                ],
+                "depth": 40.0,
+                "duration": 60.0,
+                "dt": 0.005,
+            }
+        }
+    domain_parameters_read = realisations.DomainParameters.read_from_realisation(
+        realisation_ffp
+    )
+    assert domain_parameters_read.depth == domain_parameters.depth
+    assert domain_parameters_read.duration == domain_parameters.duration
+    assert domain_parameters_read.dt == domain_parameters.dt
+    assert domain_parameters_read.resolution == domain_parameters.resolution
+    assert (
+        domain_parameters_read.domain.corners == domain_parameters.domain.corners
+    ).all()
+
+
+def test_domain_parameters_properties():
+    domain_parameters = realisations.DomainParameters(
+        resolution=0.1,  # a 0.1km resolution
+        domain=bounding_box.BoundingBox.from_centroid_bearing_extents(
+            centroid=np.array([-43.53092, 172.63701]),
+            bearing=45.0,
+            extent_x=100.0,
+            extent_y=100.0,
+        ),
+        depth=40.0,
+        duration=60.0,
+        dt=0.005,
+    )
+    assert domain_parameters.nx == 1000
+    assert domain_parameters.ny == 1000
+    assert domain_parameters.nz == 400
+
+
+def test_srf_config_example(tmp_path):
+    domain_parameters = realisations.DomainParameters(
+        resolution=0.1,  # a 0.1km resolution
+        domain=bounding_box.BoundingBox.from_centroid_bearing_extents(
+            centroid=np.array([-43.53092, 172.63701]),
+            bearing=45.0,
+            extent_x=100.0,
+            extent_y=100.0,
+        ),
+        depth=40.0,
+        duration=60.0,
+        dt=0.005,
+    )
+    srf_config = realisations.SRFConfig(
+        genslip_dt=1.0,
+        genslip_seed=1,
+        genslip_version="5.4.2",
+        resolution=0.1,
+        srfgen_seed=1,
+    )
+
+    realisation_ffp = tmp_path / "realisation.json"
+    domain_parameters.write_to_realisation(realisation_ffp)
+    srf_config.write_to_realisation(realisation_ffp)
+    with open(realisation_ffp, "r") as realisation_handle:
+        assert json.load(realisation_handle) == {
+            "domain": {
+                "resolution": 0.1,
+                "domain": [
+                    {"latitude": -43.524793866326725, "longitude": 171.76204128885567},
+                    {"latitude": -44.16756820707226, "longitude": 172.63312824122775},
+                    {"latitude": -43.53034935969409, "longitude": 173.51210368762364},
+                    {"latitude": -42.894200350955856, "longitude": 172.64076673694242},
+                ],
+                "depth": 40.0,
+                "duration": 60.0,
+                "dt": 0.005,
+            },
+            "srf": {
+                "genslip_dt": 1.0,
+                "genslip_seed": 1,
+                "resolution": 0.1,
+                "genslip_version": "5.4.2",
+                "srfgen_seed": 1,
+            },
+        }
+
+    assert realisations.SRFConfig.read_from_realisation(realisation_ffp) == srf_config
+
+
+def test_bad_domain_parameters(tmp_path: Path):
+    bad_json = tmp_path / "bad_domain_parameters.json"
+    bad_json.write_text(
+        json.dumps(
+            {
+                "domain": {
+                    "resolution": 0,  # Set to 0
+                    "domain": [
+                        {
+                            "latitude": -43.524793866326725,
+                            "longitude": 171.76204128885567,
+                        },
+                        {
+                            "latitude": -42.894200350955856,
+                            "longitude": 172.64076673694242,
+                        },
+                        {
+                            "latitude": -43.53034935969409,
+                            "longitude": 173.51210368762364,
+                        },
+                        {
+                            "latitude": -44.16756820707226,
+                            "longitude": 172.63312824122775,
+                        },
+                    ],
+                    "depth": 40.0,
+                    "duration": 60.0,
+                    "dt": 0.005,
+                }
+            }
+        )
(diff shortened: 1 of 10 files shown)


workflow/realisations.py
```diff
@@ -0,0 +1,676 @@
...
+@dataclasses.dataclass
+class HFConfig(RealisationConfiguration):
+    """High frequency simulation configuration."""
+
+    _config_key: ClassVar[str] = "hf"
+    _schema: ClassVar[Schema] = schemas.HF_CONFIG_SCHEMA
+
+    dt: float
+    """High frequency time resolution."""
+    nbu: int
```
sungeunbae, line 450: Can you copy the comments for unknown variables from below?

sungeunbae (dismissed):

workflow/realisations.py
```diff
@@ -0,0 +1,713 @@
...
+        """
+        default_config = defaults.load_defaults(defaults_version)
+        if cls._config_key not in default_config:
+            raise RealisationParseError(
+                f"No {cls._config_key} in defaults configuration"
+            )
+        return cls(**cls._schema.validate(default_config[cls._config_key]))
+
+    @classmethod
+    def read_from_realisation_or_defaults(
```
joelridden, line 154: Would we want to support cases where the realisation.json has like 1 variable defined but then the rest are grabbed from the defaults?

workflow/realisations.py
```diff
@@ -0,0 +1,713 @@
...
+
+@dataclasses.dataclass
+class DomainParameters(RealisationConfiguration):
+    """Parameters defining the spatial and temporal domain for simulation."""
+
+    _config_key: ClassVar[str] = "domain"
+    _schema: ClassVar[Schema] = schemas.DOMAIN_SCHEMA
+
+    resolution: float
+    """The simulation resoultion in kilometres."""
```
joelridden, line 342: resolution typo

workflow/schemas.py
```diff
@@ -0,0 +1,616 @@
...
+        "dyts": int,
+        "dzout": int,
+        "dzts": int,
+        "elas_only": int,
+        "enable_output_dump": int,
+        "enable_restart": int,
+        "ffault": int,
+        "fhi": float,
+        "fmax": float,
+        "fmin": float,
```
joelridden, line 493: Surely some of these can be test for is_positive or other tests etc. Even though we don't know exactly what all of these parameters do. These 2 as an example

tests/test_realisation.py
```diff
@@ -0,0 +1,403 @@
...
+                ],
+                "depth": 40.0,
+                "duration": 60.0,
+                "dt": 0.005,
+            }
+        }
+    domain_parameters_read = realisations.DomainParameters.read_from_realisation(
+        realisation_ffp
+    )
+    assert domain_parameters_read.depth == domain_parameters.depth
```
joelridden, line 49: Could these be done like the ones below and just compare the to_dict forms? Also is the name of this function right? I feel like it should be test_domain_parameters()

tests/test_realisation.py
```diff
@@ -0,0 +1,403 @@
...
+            "srf": {
+                "genslip_dt": 1.0,
+                "genslip_seed": 1,
+                "resolution": 0.1,
+                "genslip_version": "5.4.2",
+                "srfgen_seed": 1,
+            },
+        }
+
+    assert realisations.SRFConfig.read_from_realisation(realisation_ffp) == srf_config
```
joelridden, line 123: Are these 2 asserts doing the same thing?

joelridden (commented):
Just a few comments, deptry issue is interesting

tests/test_realisation.py
lispandfound: not quite, one asserts that the json is what we expect it to be, the other asserts that the deserialisation logic and schema validation works correctly.

workflow/realisations.py
lispandfound: Possibly? I haven't had the need for this yet so I haven't added it. Cesar also hasn't had a need for this just get. I'll wait and see if this becomes relevant and then add it.

workflow/schemas.py
lispandfound: Since the merge of #5 I have a better idea of what these values do so I will go through and give them sensible bounds.

tests/test_realisation.py
lispandfound: Yes I could probably refactor the test cases to use the to_dict form for most things, except for inferred properties like `nx` and `ny` which should be tested explicitly because they are not saved in the realisation and wouldn't be tested by `to_dict`.

workflow/realisations.py
joelridden: Fair call, probably will happen at some point but guess we will cross that bridge when we get there

AndrewRidden-Harper (approved):

joelridden (approved):

claudio525 (approved):

sungeunbae (approved):

lispandfound:
Closing for the same reason as https://github.com/ucgmsim/workflow/pull/8

PR #8, opened 2024-10-23
Add the Workflow Planning Script

lispandfound:
The `plan-workflow` script is the flexible workflow generator for the new workflow. It can be used to generate workflows with different goals, excluding certain stages or groups of stages, and outputting the required files and realisation contents for successful simulation.

## The Central Planning Algorithm

To allow for better critique of the code, I will explain the algorithm used to produce workflows. Example usage can be found in the test cases, which covers most of the feature set.

Let's consider the workflow stages to calculate intensity measures for a single Darfield simulation.
`plan-workflow Darfield flow.cylc --source nshm --goal im_calc`

Screenshot from 2024-10-23 15-50-05

This graph is superficially similar to the [Automated workflow pipelines](https://wiki.canterbury.ac.nz/display/QuakeCore/Automated+workflow+pipelines) page from the wiki for the old workflow.

This workflow graph represents a simulation of an NSHM rupture event "Darfield". For Cybershake, we would simply generate a flow cylc with thousands of these events and a number of realisations of each event. Researchers, however, will want to simulate custom workflows containing only part of the above graph. Cesar has run simulations with custom rupture source geometry, a custom domain, and a custom SRF. To make workflow generation as flexible as possible we need an algorithm that can skip and target parts of this graph **while still maintaining dependencies for the other stages**.

Suppose you wanted to get the simulated intensity measure calculations for a rupture where you have custom source geometry, a custom domain, and a custom SRF. The command to generate this workflow looks like:

```bash
plan-workflow Darfield flow.cylc --goal im_calc --excluding realisation_to_srf --excluding-group domain
```

In the code the planner first constructs an "abstract workflow". This workflow is the same every time you run the command. It includes all the workflow stages and dependencies.

``` mermaid
flowchart TD
    CopyInput --> NSHMToRealisation & GCMTToRealisation
    NSHMToRealisation --> SRFGeneration
    GCMTToRealisation --> SRFGeneration
    SRFGeneration --> CheckSRF
    CheckSRF --> StochGeneration
    CheckSRF --> EMOD3DParameters
    VelocityModelGeneration --> EMOD3DParameters
    VelocityModelGeneration --> HighFrequency
    StationSelection --> EMOD3DParameters
    StationSelection --> HighFrequency
    ModelCoordinates --> EMOD3DParameters
    EMOD3DParameters --> CheckDomain
    CheckDomain --> LowFrequency
    LowFrequency --> Broadband
    LowFrequency --> MergeTimeslices
    StochGeneration --> HighFrequency
    HighFrequency --> Broadband
    Broadband --> IntensityMeasureCalculation
    MergeTimeslices --> PlotTimeslices

    NSHMToRealisation --> DomainGeneration
    GCMTToRealisation --> DomainGeneration
    DomainGeneration --> EMOD3DParameters
    DomainGeneration --> VelocityModelGeneration
    DomainGeneration --> StationSelection
    DomainGeneration --> ModelCoordinates
```

This graph does not represent any real workflow; real workflows cannot have NSHM and GCMT stages for a single realisation. This workflow also includes `merge_ts` and `plot_ts`, which we don't intend to calculate. The algorithm for to generate the actual workflow proceeds as follows:

1. Firstly, transitively close this graph. By transitively close we mean that we connect every node to all its descendants directly with an edge. This will introduce a lot of redundant edges (e.g. `generate_stoch -> im_calc`), the purpose of this is to maintain the connectivity of the graph when we delete excluded nodes in the next stage.
2. We then find all the `predecessors` of every goal we specify. For our example we have just one goal, `im_calc`, but the workflow planner supports an arbitrary number of goals. The graph now looks the same minus the `merge_ts` and `plot_ts` stages. Note the flowchart below does not include the transitive edges we added in step 1.
``` mermaid
flowchart TD
    CopyInput --> NSHMToRealisation & GCMTToRealisation
    NSHMToRealisation --> SRFGeneration
    GCMTToRealisation --> SRFGeneration
    SRFGeneration --> CheckSRF
    CheckSRF --> StochGeneration
    CheckSRF --> EMOD3DParameters
    VelocityModelGeneration --> EMOD3DParameters
    VelocityModelGeneration --> HighFrequency
    StationSelection --> EMOD3DParameters
    StationSelection --> HighFrequency
    ModelCoordinates --> EMOD3DParameters
    EMOD3DParameters --> CheckDomain
    CheckDomain --> LowFrequency
    LowFrequency --> Broadband
    StochGeneration --> HighFrequency
    HighFrequency --> Broadband
    Broadband --> IntensityMeasureCalculation

    NSHMToRealisation --> DomainGeneration
    GCMTToRealisation --> DomainGeneration
    DomainGeneration --> EMOD3DParameters
    DomainGeneration --> VelocityModelGeneration
    DomainGeneration --> StationSelection
    DomainGeneration --> ModelCoordinates
```
3. The command we ran had excluded stages: the `realisation_to_srf` stage, and all the stages in the group `domain`. Additionally, because we haven't specified `--source gcmt` or `--source nshm` we are excluding `nshm_to_realisation` and `gcmt_to_realisation`. **We delete each excluded stage from the graph**
4. Finally we transitively reduce the graph again. [Transitive reduction](https://en.wikipedia.org/wiki/Transitive_reduction) is the process of removing redundant edges that were added when we transitively closed the graph in stage 1. The transitive closure and transitive reduction allows us to "heal" the workflow graph when we remove stages in step 3.

If we run the command again with the `--visualise` flag we can see the workflow graph we get at the end.

Screenshot from 2024-10-24 10-22-08

Because we did not pass the flag `--no-show-required-files` we also get a print out of all the files we require to be present for the workflow to run correctly.

```
You require the following files for your simulation:

┐
└── cylc-src
    └── WORKFLOW_NAME
        ├── flow.cylc: Your workflow file (the file expected_excluding_stages.cylc).
        └── input
            └── Darfield
                ├── Velocity_Model
                │   ├── in_basin_mask.b: Boolean basin mask (Section: Velocity Model Files).
                │   ├── rho3dfile.d: Density component file for velocity model. (Section: Velocity Model Files)
                │   ├── vp3dfile.p: p-wave velocity file for velocity model. (Section: Velocity Model Files)
                │   └── vs3dfile.s: s-wave velocity file for velocity model (Section: Velocity Model Files).
                ├── model
                │   ├── grid_file: Domain extents for EMOD3D.
                │   └── model_params: Model centre and corners for EMOD3D.
                ├── realisation.json: Realisation file for event containing: domain, metadata, velocity_model.
                ├── realisation.srf: Slip model of source (Section: SRF Format).
                └── stations
                    ├── stations.ll: lat,lon coordinates corresponding to x,y coordinates of stations in domain.
                    └── stations.statcords: x,y coordinates of stations in domain.

Refer to the indicated sections in https://wiki.canterbury.ac.nz/display/QuakeCore/File+Formats+Used+In+Ground+Motion+Simulation
Refer to the realisation glossary at URL HERE for details on filling in the realisation files.
```

So now a researcher can prepare a workflow directory with the files in the right places, and with a realisation json file containing all of the sections that will be required in the simulation.

# Plan Workflow Options

```
 Usage: plan-workflow [OPTIONS] REALISATION_IDS... FLOW_FILE

 Plan and generate a Cylc workflow file for a number of realisations.

╭─ Arguments ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    realisation_ids      REALISATION_IDS...  List of realisations to generate workflows for. Realisations have the format event:realisation_count, such as        │
│                                               Darfield:4.                                                                                                          │
│                                               [default: None]                                                                                                      │
│                                               [required]                                                                                                           │
│ *    flow_file            FILE                Path to output flow file (e.g. ~/cylc-src/my-workflow/flow.cylc) [default: None] [required]                          │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                                                                            │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                                     │
│ --help                        Show this message and exit.                                                                                                          │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Planning Workflows ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --goal                   [copy_input|gcmt_to_realisation|generate_velocity_model_parameters|  List of workflow outputs to generate [default: (dynamic)]            │
│                          generate_velocity_model|generate_station_coordinates|write_model_co                                                                       │
│                          ordinates|realisation_to_srf|check_srf|copy_domain_parameters|creat                                                                       │
│                          e_e3d_par|check_domain|generate_stoch|hf_sim|emod3d|bb_sim|im_calc|                                                                       │
│                          plot_ts|merge_ts|nshm_to_realisation]                                                                                                     │
│ --group-goal             [preprocessing|high_frequency|low_frequency|domain]                  List of group goals to generate [default: (dynamic)]                 │
│ --excluding              [copy_input|gcmt_to_realisation|generate_velocity_model_parameters|  List of stages to exclude [default: (dynamic)]                       │
│                          generate_velocity_model|generate_station_coordinates|write_model_co                                                                       │
│                          ordinates|realisation_to_srf|check_srf|copy_domain_parameters|creat                                                                       │
│                          e_e3d_par|check_domain|generate_stoch|hf_sim|emod3d|bb_sim|im_calc|                                                                       │
│                          plot_ts|merge_ts|nshm_to_realisation]                                                                                                     │
│ --excluding-group        [preprocessing|high_frequency|low_frequency|domain]                  List of stage groups to exclude [default: (dynamic)]                 │
│ --target-host            [nesi|hypocentre]                                                    Select the target host where the workflow will be run                │
│                                                                                               [default: nesi]                                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Visualising Workflows ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --visualise              --no-visualise                Visualise the planned workflow as a graph [default: no-visualise]                                           │
│ --show-required-files    --no-show-required-files      Print the expected directory tree at the start of the simulation. [default: show-required-files]            │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Sources ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --source                  [gcmt|nshm]                           If given, set the source of the realisation. For NSHM and GCMT, the realisation id corresponds to  │
│                                                                 the rupture id and GCMT PublicID respectively.                                                     │
│                                                                 [default: None]                                                                                    │
│ --defaults-version        [24.2.2.1|24.2.2.2|24.2.2.4|develop]  The simulation defaults to apply for all realisations. Required if source is specified.            │
│                                                                 [default: None]                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```


lispandfound:
Some more examples to get a feel for how the workflow generation works:

# I want to run plot ts, but I have done all the pre-processing 
`plan-workflow Darfield /tmp/expected_excluding_group.cylc --goal plot_ts  --excluding-group preprocessing`
Screenshot from 2024-10-24 10-36-46
Notice that the check stages were still run before running emod3d.
# I want to run two realisations of two GCMT events
`plan-workflow 2024p789615:2 2024p790274:2 /tmp/expected_excluding_group.cylc --goal im_calc --source gcmt  --defaults-version 24.2.2.2`
image
Notice that for events with multiple realisations we generate the velocity model just once and share it with the other realiasation.
# As above, but now I already have done all the pre-processing
`plan-workflow 2024p789615:2 2024p790274:2 /tmp/expected_excluding_group.cylc --goal im_calc --source gcmt  --defaults-version 24.2.2.2 --excluding-group preprocessing`
image
The realisations are now independent because the velocity model is already generated.

lispandfound:
As one last example I'd like to illustrate how we would use this tool for Cybershake. We would actually plan two workflows

1. A pre-processing workflow, which is run on hypocentre (or the upcoming RCC cluster) for validation. We can just target the preprocessing group (which is an alias for all of the terminal stages of preprocessing)
`plan-workflow Rupture_0:10 Rupture_1:10 ... (remaining nshm ruptures, probably read from a file with xargs) flow.cylc --group-goal preprocessing --source nshm --target-host hypocentre`
2. Create a second workflow intended to be run on NeSI for actual simulation and IM Calculation. This would be done after validating the results of the previous workflow on hypocentre.
`plan-workflow Rupture_0:10 Rupture_1:10 ... (remaining nshm ruptures, probably read from a file with xargs) flow.cylc --excluding-group preprocessing --goal im_calc`

sungeunbae:
> 2\. We then find all the `predecessors` of every goal we specify. For our example we have just one goal, `im_calc`, but the workflow planner supports an arbitrary number of goals. The graph now looks the same minus the `merge_ts` and `plot_ts` stages. Note the flowchart below does not include the transitive edges we added in step 1.
> 
> ```mermaid
> flowchart TD
>     CopyInput --> NSHMToRealisation & GCMTToRealisation
>     NSHMToRealisation --> SRFGeneration
>     GCMTToRealisation --> SRFGeneration
>     SRFGeneration --> CheckSRF
>     CheckSRF --> StochGeneration
>     CheckSRF --> EMOD3DParameters
>     VelocityModelGeneration --> EMOD3DParameters
>     VelocityModelGeneration --> HighFrequency
>     StationSelection --> EMOD3DParameters
>     StationSelection --> HighFrequency
>     ModelCoordinates --> EMOD3DParameters
>     EMOD3DParameters --> CheckDomain
>     CheckDomain --> LowFrequency
>     LowFrequency --> Broadband
>     StochGeneration --> HighFrequency
>     HighFrequency --> Broadband
>     Broadband --> IntensityMeasureCalculation
> 
>     NSHMToRealisation --> DomainGeneration
>     GCMTToRealisation --> DomainGeneration
>     DomainGeneration --> EMOD3DParameters
>     DomainGeneration --> VelocityModelGeneration
>     DomainGeneration --> StationSelection
>     DomainGeneration --> ModelCoordinates
> ```

sungeunbae:
Impressive quality of PR documentation. Worth a preservation as a proper page

lispandfound:
> `merge_ts` stage still appears in your diagram.
> 
> > 2. We then find all the `predecessors` of every goal we specify. For our example we have just one goal, `im_calc`, but the workflow planner supports an arbitrary number of goals. The graph now looks the same minus the **`merge_ts` and `plot_ts`** stages. Note the flowchart below does not include the transitive edges we added in step 1.

Oh good point, I forgot to remove it in the markdown.

sungeunbae (commented):
I like the idea of using `runner_latest.sif`, assuming it is the latest image.
How do you manage the image version coupled with the computation version (eg. v24_2_2_1)? Considering the computation version is intended to replicate the method and reproduce the same result, I guess the record of the image version should be kept somewhere?

lispandfound:
> I like the idea of using `runner_latest.sif`, assuming it is the latest image.
> How do you manage the image version coupled with the computation version (eg. v24_2_2_1)? Considering the computation version is intended to replicate the method and reproduce the same result, I guess the record of the image version should be kept somewhere?

So I have already thought about this a little bit and I think I have come up with a solution. The situation we're trying to avoid is one where we run a simulation with some defaults, and then later cannot reproduce the outputs because some of these values are unknown. To do this, for any stage that reads a section with defaults (let's use the velocity model section for example), the following occurs:

1. It first checks that the realisation does not already contain a section "velocity_model". If it does it reads this section first and does not look at the defaults.
2. If the realisation does not already contain a section "velocity_model", it then reads it from the defaults yaml file.
3. If the defaults are loaded, it writes these defaults to the realisation.

Because of step 1, rerunning a realisation two years later will result in the same output because the realisation's values for the section would be read first - not the containers. Because of step 3 the default values are copied into the realisation at the first run time, so you get a recording of every parameter the simulation was run with.

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+            print("You require the following files for your simulation:")
+            print()
+            printree.ptree(missing_file_tree)
+            print()
+            print(
+                "Refer to the indicated sections in https://wiki.canterbury.ac.nz/display/QuakeCore/File+Formats+Used+In+Ground+Motion+Simulation"
+            )
+        if any(required_realisation_sections.values()):
+            print(
+                "Refer to the realisation glossary at URL HERE for details on filling in the realisation files."
```
AndrewRidden-Harper, line 1037: Do you want to add this URL before merging?

AndrewRidden-Harper (approved):
Very impressive!

sungeunbae:
> So I have already thought about this a little bit and I think I have come up with a solution. The situation we're trying to avoid is one where we run a simulation with some defaults, and then later cannot reproduce the outputs because some of these values are unknown. 

I'm talking about something slightly different. The situation I want to avoid is one where the software inside "runner_latest.sif" changes behaviour. Each time we make a new version of container, it will include many code changes. Let's say we produced some sim data in 2024 using the computation version 24.2.2.1 with the "runner_latest.sif". In 2026, we hope to reproduce the same data using the same version 24.2.2.1, but we only know we used "runner_latest.sif" in 2024, which may not be the same thing as "runner_latest.sif" in 2026.  I'm not sure if we keep the record of container version somewhere..?

lispandfound:
> I'm talking about something slightly different. The situation I want to avoid is one where the software inside "runner_latest.sif" changes behaviour. Each time we make a new version of container, it will include many code changes. Let's say we produced some sim data in 2024 using the computation version 24.2.2.1 with the "runner_latest.sif". In 2026, we hope to reproduce the same data using the same version 24.2.2.1, but we only know we used "runner_latest.sif" in 2024, which may not be the same thing as "runner_latest.sif" in 2026. I'm not sure if we keep the record of container version somewhere..?

So I think the solution to this is human rather than software. The container is just a file and can be archived with each Cybershake run we do. I would do this along with the cylc flow file. If we need to reproduce the run, the container can be downloaded and rerun with that container.

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade coverage[toml]
+
+      - name: Download coverage data
+        uses: actions/download-artifact@v4
+        with:
+          name: coverage-data
+
+      - name: Combine coverage and fail it it’s under 100 %
```
joelridden, line 58: 95%?

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+
+
+class WorkflowTarget(StrEnum):
+    """Enumeration of possible workflow targets."""
+
+    NeSI = "nesi"
+    Hypocentre = "hypocentre"
+
+
+class StageIdentifier(StrEnum):
```
joelridden, line 33: I feel like these Enums can be quite useful and potentially importable from elsewhere in future, maybe having these in a file that isnt a script to import from could be good?

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+    Adds all stages for the realisation to run, and links to event
+    stages for shared resources (i.e. the velocity model).
+
+    Parameters
+    ----------
+    workflow_plan : nx.DiGraph
+        The current workflow paln.
+    event : str
+        The event to add.
+    sample : Optional[int]
```
joelridden, line 593: is rel_num or something similar slightly better here rather than sample?

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+
+    excluding_stages = {
+        Stage(excluded, *realisation)
+        for excluded in excluding
+        for realisation in realisations
+    }
+
+    output_graph = nx.DiGraph()
+    realisation_iteration = (
+        realisations if len(realisations) < 100 else tqdm.tqdm(realisations)
```
joelridden, line 729: could tqdm just always be used?

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+                    )
+                    - excluding_stages
+                )
+            )
+            output_graph.update(
+                edges=reduced_graph.edges(), nodes=reduced_graph.nodes()
+            )
+
+    roots = [node for node, degree in output_graph.in_degree() if degree == 0]
+    copy_input_stage = Stage(StageIdentifier.CopyInput, "", None)
```
joelridden, line 751: Can the user select to remove CopyInput in the excluded stages? Just wondering what weird issues you might get if that's the case, it might need to be checked at the script execution just incase.

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+                ),
+                default=0,
+            ),
+        )
+    for stage, next_stage in workflow_plan.edges():
+        network.add_edge(stage_to_node_string(stage), stage_to_node_string(next_stage))
+    return network
+
+
+REALISATION_ITERATION_RE = r"_rel\d+$"
```
joelridden, line 821: Should this be at the top?

workflow/scripts/plan_workflow.py
```diff
@@ -0,0 +1,1042 @@
...
+    flow_file: Annotated[
+        Path,
+        typer.Argument(
+            help="Path to output flow file (e.g. ~/cylc-src/my-workflow/flow.cylc)",
+            writable=True,
+            dir_okay=False,
+        ),
+    ],
+    goal: Annotated[
+        list[StageIdentifier],
```
joelridden, line 891: Edge case, what happens when the goal is CopyInput?

.github/workflows/pytest.yml
lispandfound: yes that's a fair point.

workflow/scripts/plan_workflow.py
lispandfound: copy input cannot be excluded from the run. It will not crash if the input directory doesn't exist, if it doesn't it will exit without running.

sungeunbae (approved):

lispandfound:
I'm going to close this PR, pursuant to the new PR strategy we intend to adopt for the new workflow.

PR #9, opened 2025-01-13
Point to the main branch of velocity_modelling

lispandfound:
Just updates the requirements.txt file to use the main branch of the velocity_modelling repository now that https://github.com/ucgmsim/velocity_modelling/pull/6 has been merged.


.github/workflows/pytest.yml
joelridden, line 1: Why was this removed?

lispandfound: because the tests don't work and haven't for a while. They should be updated, but I can't see the point at this stage because we don't have a database of known-good outputs to compare to yet. The other problem is that they take too long on GitHub runners (I think they need to run on Hypocentre). I think whole workflow should be re-introduced at a later date.

lispandfound: It takes a ton of engineering time to set up the tests and they can at most tell you pass/fail, but not why. With most of the scientific code being unit tested elsewhere, the tests serve only to check that the parts that call out to Rob's code work correctly.

joelridden (approved):

PR #12, opened 2025-01-14
Scientific Registry and Cleanup

lispandfound:
This PR integrates the scientific registry in https://github.com/ucgmsim/qcore/pull/341 with the workflow. It replaces the old scientific defaults with defaults sourced from the registry. It also takes the opportunity to remove and cleanup the directory structure.


workflow/scripts/check_srf.py
```diff
@@ -24,6 +24,12 @@ def check_srf(
         Path, typer.Argument(help="The path to the realisation for the SRF.")
     ],
     srf_ffp: Annotated[Path, typer.Argument(help="The path to the SRF file to check.")],
+    velocity_model_path: Annotated[
```
joelridden: Just a potential rename of registry_velocity_model_path to remove some potential confusion. Along with all the other ones

claudio525 (dismissed):

joelridden (dismissed):

AndrewRidden-Harper (dismissed):

workflow/realisations.py
```diff
@@ -22,6 +22,7 @@
 from schema import Schema
 from velocity_modelling.bounding_box import BoundingBox
 
+from IM import im_calculation
```
claudio525, line 25: Whats IM? Its not a package and I can't see a module with that name?

claudio525: I think its meant to be
```suggestion
from im_calculcation import IM
```

lispandfound: nope, the IM package is the actual import we want.

lispandfound: It isn't called im_calculation for historical reasons... I should've changed that in the IM calc repo rewrite but I didn't.

PR #13, opened 2025-01-21
Add rupture velocity delay

lispandfound:
This PR refactors `realisation-to-srf` SRF stitching, and adds rupture jumping delay. Also fixes a bug related to parent SRF jumping using the wrong fault to compute the parent point.

From what I can see from the addition of rupture front speed into the delay does not make a large difference.


sungeunbae (approved):

AndrewRidden-Harper (approved):
looks good

PR #19, opened 2025-02-19
Use the new qcore CLI module + unit tests

lispandfound:
Implements a refactor of the CLI scripts to use the new `qcore.cli` module implemented in https://github.com/ucgmsim/qcore/pull/344, and then reintroduces unittesting for the workflow module. The unit tests have coverage standards, but the coverage ignores the vast majority of the codebase (in scripts). This is because these scripts will be tested properly with E2E testing on mantle, and I hope to somehow have coverage metrics for those files from the E2E testing flow. The unittests run by the github action will test the rest of the codebase, plus run `<workflow-stage> --help` for each workflow stage to ensure that the scripts will still at least not fail immediately. These checks are quick to unit test so that you can get feedback and quickly fix something without waiting hours for the E2E tests to execute on the workflow.


lispandfound:
Tests will fail until https://github.com/ucgmsim/qcore/pull/344 merges.

workflow/scripts/create_e3d_par.py
```diff
@@ -224,66 +227,18 @@ def format_as_emod3d_value(value: int | float | str | Path) -> str:
...
-    emod3d_version: Annotated[
-        str, typer.Option(help="The version of the EMOD3D binary to use.")
-    ] = "3.0.8",
+    realisation_ffp: Annotated[Path, typer.Argument(exists=True, readable=True, dir_okay=False)],
+    srf_file_ffp: Annotated[Path, typer.Argument(exists=True, readable=True, dir_okay=False)],
+    velocity_model_ffp: Annotated[Path, typer.Argument(exists=True, readable=True, file_okay=False)],
+    stations_ffp: Annotated[Path, typer.Argument(exists=True, readable=True, file_okay=False)],
+    grid_ffp: Annotated[Path, typer.Argument(exists=True, readable=True, file_okay=False)],
+    output_ffp: Annotated[Path, typer.Argument(writable=True, file_okay=False)],
+    scratch_ffp: Annotated[Path, typer.Option(writable=True, file_okay=False)] = Path("/out"),
```
sungeunbae, line 238: Screenshot 2025-02-20 at 10 23 15 AM

Does the top-level docstring need an update?

lispandfound: Yes it does

lispandfound: Actually no it doesn't? The arguments are accurate. Options are swept under the `[OPTIONS]` block. If you want detailed explanation, then as the next paragraph suggests you should run `create-e3d-par --help`

joelridden (approved):

claudio525 (approved):

AndrewRidden-Harper (approved):

PR #23, opened 2025-02-25
Add More SRF Checks

lispandfound:
After the discovery of Cesear's SRF bug, I thought it prudent to introduce more SRF checks to catch future bugs like this before they reach EMOD3D. Here are the additional quick checks I've added: 

1. Check that there are no NaN values in the SRF.
2. Check that `-90 <= lat <= 90` and `-180 <= lon <= 180` for all points in the srf.
3. Check that the original SRF geometry can be extracted.

The last check invokes the `.planes` property to which constructs source modelling `Plane` objects. The `Plane` constructor checks that source modelling does apply here. This means SRFs meeting the above three criteria, but specifying a group of points that *do not* form a plane will be rejected.

Additionally, I have fixed the SRF magnitude check. This check ensures that the magnitude in the SRF matches the magnitude in the realisation file. This was broken because I wasn't calculating SRF magnitude correctly due to a simple mistake. I have corrected this mistake, and testing the ~100 alpine realisation SRFs generated for Brendon a few weeks ago I now get passes for all of them.

### DIFF
--- a/workflow/log_utils.py
+++ b/workflow/log_utils.py
@@ -1,8 +1,8 @@
 r"""Structured Logging Utilities for `logging`.
 
-This module support for different structured-logging formats (JSON and
+This module supports different structured-logging formats (JSON and
 text) and decorators for logging function calls and command
-executions. The logging functions can capture and structure log
+executions. The logging functions capture and structure log
 messages along with additional contextual information such as function
 arguments, execution status, and timestamps.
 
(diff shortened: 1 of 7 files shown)


joelridden (approved):

AndrewRidden-Harper (approved):

joelridden (approved):

AndrewRidden-Harper (approved):

lispandfound:
Bypassed merge rules because they appeared not to work.

PR #38, opened 2025-04-27
Support PyNZVM for velocity model generation

lispandfound:
This small PR adds a flag `--use-pynzvm`. When enabled, it imports pynzcvm and executes the main entry point to the script. This is disabled by default.

### DIFF
--- a/workflow/scripts/generate_velocity_model.py
+++ b/workflow/scripts/generate_velocity_model.py
@@ -123,13 +123,13 @@ def run_nzvm(
     )
 
 
-def run_pynzvm(nzvm_config_ffp: Path) -> None:
-    """Run PyNZVM executable with specified configuration.
+def run_nzcvm(nzvm_config_ffp: Path) -> None:
+    """Run NZCVM executable with specified configuration.
 
     Parameters
     ----------
     nzvm_config_ffp : Path
-        Path to the NZVM configuration file.
+        Path to the NZVM-format configuration file.
     """
     from velocity_modelling.scripts import nzcvm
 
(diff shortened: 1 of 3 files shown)


joelridden (approved):

workflow/scripts/generate_velocity_model.py
```diff
@@ -123,13 +123,35 @@ def run_nzvm(
     )
 
 
+def run_pynzvm(nzvm_config_ffp: Path) -> None:
```
sungeunbae: We had multiple names that floated around during the development, and I ended up converging to nzcvm.  For many people, they don't care if it is C or Python, but what matters is the first "community" focused VM.

lispandfound: Sure.

joelridden (approved):

workflow/scripts/generate_velocity_model.py
```diff
@@ -149,6 +171,8 @@ def generate_velocity_model(
         Path to the NZVM binary.
     work_directory : Path, optional
         Directory for intermediate output files.
+    use_nzcvm : bool, optional
+        If True, use the Nzcvm Python package instead of the NZVM binary. Default is False.
```
AndrewRidden-Harper:
```suggestion
        If True, use the NZCVM Python package instead of the NZVM binary. Default is False.
```

AndrewRidden-Harper (commented):
small suggested change to use NZCVM in a comment

sungeunbae (approved):
Just fix as per Andrew's suggestion, I'm happy.

joelridden (approved):

AndrewRidden-Harper (approved):

PR #43, opened 2025-05-13
Add type annotations for all returns and enable ruff annotation checks

lispandfound:
This PR adds almost all annotation type checking to the project and then adds type annotations where ruff requires it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,13 @@ extend-select = [
     "UP035",
     # Missing function argument type-annotation
     "ANN001",
+    "ANN002",
+    "ANN003",
+    "ANN201",
+    "ANN202",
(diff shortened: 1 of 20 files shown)


lispandfound:
Tests will fail until: https://github.com/fastapi/typer/pull/1145 merges. Turns out the problem is `click`, a dependency of typer. So downgrading typer won't fix this issue, we need something like `click < 0.8.2` to the project and then adding a deptry exclusion for `click`. This is, of course, extremely annoying and I'd prefer that typer just fixed their repo first. Researchers have already noticed this problem so I don't know how long we have to wait.

joelridden (approved):

AndrewRidden-Harper (approved):

PR #44, opened 2025-05-15
Extract rakes and magnitudes out of rupture propagation

lispandfound:
An annoying but necessary change. The rupture propagation config included magnitudes and rakes, two things that really aren't related to rupture propagation trees or jump points.
This PR refactors the object all downstream workflow stages to use separate `Rakes` and `Magnitudes` sections. It is the difference between a realisation containing

``` json
{
  "rupture_propagation": {
    "rupture_causality_tree": {
      "fault": null
    },
    "magnitudes": {
      "fault": 5.5
    },
    "rakes": {
      "fault": 75
    },
    "jump_points": {},
    "hypocentres": {
      "fault": {
        "s": 0.5,
        "d": 0.5
      }
    }
  }
}
```
and

``` json
{
  "rupture_propagation": {
    "rupture_causality_tree": {
      "fault": null
    },
    "jump_points": {},
    "hypocentres": {
      "fault": {
        "s": 0.5,
        "d": 0.5
      }
    }
  },
  "magnitudes": {
    "magnitudes": {
      "fault": 5.5
    }
  },
  "rakes": {
    "rakes": {
      "fault": 75
    }
  }
}
```

The end result is that a better separation of concerns is obtained and various stages have simplified logic and interfaces because of that, particular `generate-rupture-propagation`.


workflow/scripts/generate_rupture_propagation.py
```diff
@@ -40,101 +40,17 @@
...
-from qcore.uncertainties import distributions, mag_scaling
+from qcore.uncertainties import distributions
 from source_modelling import rupture_propagation
-from source_modelling.sources import Fault
 from workflow import realisations
 
 app = typer.Typer()
 
 
-def a_to_mw_leonard(area: float, rake: float) -> float:
```
joelridden, line 51: Are all these functions not needed now?

lispandfound: They aren't needed in `generate-rupture-propagation` because that function no longer needs to supply magnitudes. This is duplicated code from other stages, precisely because the rupture propagation config had too much in it. So the code is no longer needed.

joelridden (approved):
Just need to deal with merge conflicts

AndrewRidden-Harper (approved):

sungeunbae (approved):

joelridden (approved):

AndrewRidden-Harper (approved):

PR #46, opened 2025-05-15
Add container build action

lispandfound:
Adds a container build action that automatically uploads containers to dropbox when:

1. A PR merges to main or pegasus, and
2. The tests pass.

This should simplify the deployment of fixes for researchers, we can just point them to a link they can download.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


lispandfound:
I'm going to merge this bypassing checks because it doesn't affect any code. Will create a PR for fixes if needed. Main reason is I cannot test it without first having the action in the source tree...

PR #50, opened 2025-05-19
Make the HF 1D velocity model the standard and delete the HF version

AndrewRidden-Harper:
This PR adopts the HF 1D velocity model Vs = 0.5 values in the first two layers, and the HF thickness of 999.9999 in the last layer to resolve discrepancies. Now there is no need for a separate HF 1D velocity model, so it has been deleted.


lispandfound (approved):

joelridden (approved):

lispandfound:
@AndrewRidden-Harper You'll have to merge from `pegasus` first.

lispandfound (approved):
LGTM!

PR #51, opened 2025-05-20
pass fault bottom instead of hypocentre to get_max_depth() function

AndrewRidden-Harper:
This PR changes how `max_depth` is calculated. Now the `get_max_depth()` function is given the fault's bottom instead of hypocentre, making the `max_depth` lower


AndrewRidden-Harper:
Changed it to use initial_fault.planes[0].bottom_m as @lispandfound has already implemented that

lispandfound (approved):

lispandfound:
@joelridden or @claudio525 could you please review this to move it along? Cesar requires it for his simulations.

joelridden (approved):

PR #52, opened 2025-05-20
Makes workflow nz rounding consistent with C NZVM code

AndrewRidden-Harper:
This PR makes the workflow round the number of z coordinate positions in the discretised domain (nz) in the same way as the C NZVM code, which just truncates the nz value

### DIFF
--- a/workflow/realisations.py
+++ b/workflow/realisations.py
@@ -445,19 +445,22 @@ class DomainParameters(RealisationConfiguration):
     @property
     def nx(self) -> int:  # numpydoc ignore=RT01
         """int: The number of x coordinate positions in the discretised domain."""
-        return int(np.round(self.domain.extent_x / self.resolution))
+        # The C NZVM code truncates (removes the decimal part) of the nx value,
+        # so we do the same for consistency
+        return int(self.domain.extent_x / self.resolution)
 
     @property
     def ny(self) -> int:  # numpydoc ignore=RT01
         """int: The number of y coordinate positions in the discretised domain."""
-        return int(np.round(self.domain.extent_y / self.resolution))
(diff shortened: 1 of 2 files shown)


workflow/realisations.py
```diff
@@ -457,7 +455,10 @@ def ny(self) -> int:  # numpydoc ignore=RT01
     @property
     def nz(self) -> int:  # numpydoc ignore=RT01
         """int: The number of z coordinate positions in the discretised domain."""
-        return int(np.round(self.depth / self.resolution))
+        # The C NZVM code truncates (removes the decimal part) of the nz value,
+        # so we do the same for consistency. The C NZVM code does round the nx and ny
+        # values, so it is unclear why the nz value is treated differently.
+        return int(self.depth / self.resolution)
```
lispandfound: Allegedly the C code also does this for `nx` and `ny`. Please update those to match the `nz` calculations. This was discovered by @sungeunbae.

lispandfound (changes requested):

AndrewRidden-Harper:
nx and ny were also inconsistent with the C NZVM code's use of truncation instead of rounding, so they have also been updated in the same way as for nz

AndrewRidden-Harper:
It should now replicate the behaviour of C always rounding 0.5 up

lispandfound (approved):

AndrewRidden-Harper:
With this rounding of int(x+0.5), I had to revert test_domain_parameters_properties() to look for nx = 1000, and ny = 1000 (instead of 999)

lispandfound (approved):

joelridden (approved):

sungeunbae (approved):

PR #56, opened 2025-07-07
Adds support for using a point source approximation in the new workflow

AndrewRidden-Harper:
This PR adds support for using a point source approximation in the new workflow. The figure below verifies our implementation of the magnitude-slip relation
new_workflow_magnitude_slip_plot


lispandfound:
@AndrewRidden-Harper please address the pytest checks first.

workflow/scripts/realisation_to_srf.py
```diff
@@ -554,18 +726,33 @@
...
+                srf_name,
+            )
+
+            shutil.copyfile(work_directory / (srf_name + ".srf"), output_srf_filepath)
+
+        else:
+            generate_point_source_srf(name, params, environment, output_srf_filepath)
+
+        realisations.append_log_entry(realisation_ffp)
+        srf_config.write_to_realisation(realisation_ffp)
```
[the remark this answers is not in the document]
AndrewRidden-Harper: @lispandfound Am I correct in thinking that we wouldn't ever have a situation of realisations with multiple source geometries

AndrewRidden-Harper:
@lispandfound It's passing the tests. Gemini made some comments but I don't know if we need to address them so let me know what you think

workflow/realisations.py
```diff
@@ -363,6 +363,22 @@ class SRFConfig(RealisationConfiguration):
     resolution: float
     """The resolution of the SRF geometry"""
 
+    # The following attributes are only used for the point source approximation
+    stype: Optional[str] = None
+    """Slip type for generic_slip2srf"""
```
lispandfound: Make `stype` an Enum so that invalid values can't be passed.

workflow/scripts/gcmt_to_realisation.py
```diff
@@ -190,33 +200,62 @@ def gcmt_to_realisation(
...
 
-    centroid = np.array([latitude, longitude, centroid_depth])
-    plane = sources.Plane.from_centroid_strike_dip(
-        centroid,
-        selected_nodal_plane.dip,
-        length,
-        width,
-        strike=selected_nodal_plane.strike,
-    )
+        length_m = np.sqrt(area_km2) * 1000  # Convert km to meters
```
lispandfound: Check with Felipe if this should always be a square, or if the length should be sampled from the leonard proportional scaling relationship.

workflow/scripts/realisation_to_srf.py
```diff
@@ -502,6 +522,162 @@ def generate_fault_srfs_multi(
...
+    Returns
+    -------
+    float
+        The calculated slip in cm.
+    """
+
+    # The factor of 1.0e-20 converts the combination of input units to cm.
+    return (moment_dyne_cm * 1.0e-20) / (
+        fault_area_km2 * rho_g_per_cm3 * vs_km_per_s**2
+    )
```
lispandfound: It is probably overkill to introduce another function for this IMO

```suggestion
    # The factor of 1.0e-20 converts the combination of input units to cm.
    return (moment_dyne_cm * 1.0e-20) / (
        fault_area_km2 * rho_g_per_cm3 * vs_km_per_s**2
    )
```

workflow/scripts/realisation_to_srf.py
```diff
@@ -502,6 +522,162 @@ def generate_fault_srfs_multi(
             )
 
 
+def calc_point_source_slip_wrapper(params: SRFRealisationContext, name: str) -> float:
```
lispandfound:
```suggestion
def calc_point_source_slip(params: SRFRealisationContext, name: str) -> float:
```

see below

lispandfound: Also rather than passing in the whole context, which I really only added for the big multiprocessing wrapper which needs it, just pass the velocity model, fault, and magnitude.

lispandfound: These sorts of scientific calculations I usually want to have in source modelling, so actually I think it should go there, probably in the `moment` module.

lispandfound (changes requested):

workflow/scripts/realisation_to_srf.py
```diff
@@ -118,6 +125,10 @@ def generate_fault_gsf(
     gsf_output_filepath = gsf_output_directory / f"{name}.gsf"
     gsf_df = gsf.source_to_gsf_dataframe(geometry, subdivision_resolution)
     gsf_df["loc_rake"] = rake
+    if slip is not None:
+        gsf_df["slip"] = [slip]
```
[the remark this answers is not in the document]
AndrewRidden-Harper: Changed

workflow/scripts/realisation_to_srf.py
```diff
@@ -554,18 +726,33 @@ def generate_srf(
...
-    shutil.copyfile(work_directory / (srf_name + ".srf"), output_srf_filepath)
-    realisations.append_log_entry(realisation_ffp)
+    for name, geometry in params.source_config.source_geometries.items():
+        if isinstance(geometry, Fault):
+            generate_fault_srfs_parallel(
+                source_config.source_geometries, params, environment
+            )
+
+            stitch_srf_files(
+                source_config.source_geometries,
```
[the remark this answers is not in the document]
AndrewRidden-Harper: source_config is defined in this scope on line 738

workflow/scripts/gcmt_to_realisation.py
AndrewRidden-Harper: added a warning following Brendon's [suggestion](https://uceqeng.slack.com/archives/C03V84ZD21H/p1752041664196469?thread_ts=1747609850.094969&cid=C03V84ZD21H)

workflow/realisations.py
```diff
@@ -33,6 +34,23 @@
...
+    ucsb = "ucsb"
+    ucsb2 = "ucsb2"
+    ucsb_T = "ucsb-T"  # noqa: N815
+    ucsb_varT1 = "ucsb-varT1"  # noqa: N815
+    cos = "cos"
+    seki = "seki"
+
+
+# Map the string representation of the stype to the Stype enum.
+stype_enum_mapping = {member.value: member for member in Stype}
```
lispandfound: This is unnecessary.

workflow/schemas.py
```diff
@@ -296,6 +298,32 @@ def _corners_to_array(corners_spec: list[dict[str, float]]) -> np.ndarray:
         Literal("resolution", description="Subdivision resolution."): And(
             float, _is_positive
         ),
+        Literal("stype", description="Slip type for generic_slip2srf"): Or(
+            None, 
+            And(str, lambda s: s in ALLOWED_STYPE_VALUES)
```
lispandfound: You should actually convert this to the correct type:

```suggestion
            Use(Stype)
```

This has the positive side-effect of meaning you don't need the dictionary to check if it is an allowed value, because `Stype(string)` will raise an exception if `string` is not the name of one of its members.

workflow/scripts/gcmt_to_realisation.py
lispandfound: `The more detailed approach is the aspect ratio scaling details in Leonard etc, you could note this in the code and that we will implement it at a later date.`

We have actually already implemented these details. See [magnitude_to_length_width](https://quakecoresoft.canterbury.ac.nz/docs/source_modelling.magnitude_scaling.html#magnitude_to_length_width) which samples out of the leonard scaling relationships to produce lengths/widths consistent with the area relationships.

lispandfound: Hence, you should use that function to implement the dimensions.

workflow/scripts/realisation_to_srf.py
```diff
@@ -552,23 +676,34 @@ def generate_srf(
...
+                srf_name,
+            )
+
+            shutil.copyfile(work_directory / (srf_name + ".srf"), output_srf_filepath)
+
+        else:
+            generate_point_source_srf(name, params, environment, output_srf_filepath)
+
+        realisations.append_log_entry(realisation_ffp)
+        srf_config.write_to_realisation(realisation_ffp)
```
lispandfound: This code is still broken! It should remain unchanged from the original.
```suggestion
    generate_fault_srfs_multi(
        source_config.source_geometries,
        params,
        environment,
        single_threaded=single_threaded,
    )
    srf_name = normalise_name(metadata.name)
    stitch_srf_files(
        source_config.source_geometries,
        rupture_propagation,
        work_directory,
        srf_name,
    )
    srf_config.write_to_realisation(realisation_ffp)

    shutil.copyfile(work_directory / (srf_name + ".srf"), output_srf_filepath)
    realisations.append_log_entry(realisation_ffp)
```

Instead `generate_fault_srfs_multi` needs to be modified to generate point source SRFs.

lispandfound (changes requested):

workflow/scripts/gcmt_to_realisation.py
AndrewRidden-Harper: @lispandfound Thanks for pointing out [magnitude_to_length_width](https://quakecoresoft.canterbury.ac.nz/docs/source_modelling.magnitude_scaling.html#magnitude_to_length_width). Point sources only have length_m, so am I correct in thinking that we would need to change to using a Plane source to use separate length_m and width_m?

workflow/realisations.py
```diff
@@ -363,6 +364,22 @@ class SRFConfig(RealisationConfiguration):
...
+    """Rise time for generic_slip2srf"""
+
+    risetimefac: Optional[float] = None
+    """Rise time factor for generic_slip2srf"""
+
+    risetimedep: Optional[float] = None
+    """Rise time depth dependency for generic_slip2srf"""
+
+    inittime: Optional[float] = None
+    """Initial time for generic_slip2srf"""
```
lispandfound: Apparently `Optional` is deprecated now? Or, at least, ruff keeps nagging me about it. Better to use 


```suggestion
    stype: schemas.Stype | None = None
    """Slip type for generic_slip2srf"""

    risetime: float | None = None
    """Rise time for generic_slip2srf"""

    risetimefac: float | None = None
    """Rise time factor for generic_slip2srf"""

    risetimedep: float | None = None
    """Rise time depth dependency for generic_slip2srf"""

    inittime: float | None = None
    """Initial time for generic_slip2srf"""
```
for new code at any rate.

workflow/scripts/realisation_to_srf.py
```diff
@@ -491,6 +522,96 @@ def generate_fault_srfs_multi(
...
+
+    environment.gsf_directory.mkdir(parents=True, exist_ok=True)
+
+    fault = params.source_config.source_geometries[name]
+
+    resolution = params.srf_config.resolution
+
+    # Get magnitude and convert to seismic moment
+    magnitude = params.magnitudes.magnitudes[name]
+    moment_dyne_cm = mag_scaling.mag2mom(magnitude)
```
lispandfound: Once you've added magnitude to moment, use that in place of mag2mom. That should output Nm, and then the conversion is unnecessary.

workflow/scripts/realisation_to_srf.py
```diff
@@ -562,12 +693,23 @@ def generate_srf(
...
+            source_config.source_geometries,
+            rupture_propagation,
+            work_directory,
+            srf_name,
+        )
+    else:
+        shutil.copyfile(
+            work_directory / "srf" / (srf_name + ".srf"),
+            work_directory / (srf_name + ".srf"),
+        )
```
lispandfound: seems like a bug in `stitch_srf_files` if it is changing the slip value.

lispandfound (changes requested):

workflow/scripts/realisation_to_srf.py
```diff
@@ -478,9 +499,20 @@ def generate_fault_srfs_multi(
...
     environment.gsf_directory.mkdir(exist_ok=True)
     params.velocity_model_1d.write_velocity_model(environment.velocity_model_path)
+
+    # In practice, a Point source will be the only item in `faults`, so if there
+    # is a Point source, restrict to `single_threaded` mode for simplicity.
+    for name in faults:
+        if isinstance(faults[name], Point):
+            single_threaded = True
+            break
+
```
lispandfound: I don't like this check because it would be better if it just checked for a single fault instead 
```suggestion
    single_threaded = single_threaded or len(faults) == 1
```

workflow/scripts/realisation_to_srf.py
```diff
@@ -478,9 +499,20 @@ def generate_fault_srfs_multi(
...
     if single_threaded:
         for name in faults:
-            generate_fault_srf(name, params, environment)
+            if isinstance(faults[name], Point):
+                generate_point_source_srf(name, params, environment)
+            else:
+                generate_fault_srf(name, params, environment)
```
lispandfound: I would prefer if `generate_fault_srf` dispatched to either:

1. generate_point_source_srf
2. Something like generate_finite_fault_srf.

Again, just moving the special casing down the call-stack will improve code reuse here imo.

lispandfound: That is the if statement should be in `generate_fault_srf`

workflow/scripts/realisation_to_srf.py
```diff
@@ -502,35 +629,42 @@ def generate_srf(
         "/out"
     ),
     genslip_path: Annotated[Path, typer.Option(readable=True, dir_okay=False)] = Path(
-        "/EMOD3D/tools/genslip_v5.4.2"
+        "/home/arr65/src/EMOD3D/tools/genslip_v5.4.2"
     ),
+    generic_slip2srf_path: Annotated[
+        Path, typer.Option(readable=True, dir_okay=False)
+    ] = Path("/home/arr65/src/EMOD3D/tools/generic_slip2srf"),
```
lispandfound: These paths are wrong.

lispandfound (changes requested):

workflow/scripts/realisation_to_srf.py
AndrewRidden-Harper: @lispandfound I've moved this check into `generate_fault_srf`. Let me know if I've misinterpreted

workflow/scripts/realisation_to_srf.py
```diff
@@ -502,35 +628,42 @@ def generate_srf(
         "/out"
     ),
     genslip_path: Annotated[Path, typer.Option(readable=True, dir_okay=False)] = Path(
-        "/EMOD3D/tools/genslip_v5.4.2"
+        "EMOD3D/tools/genslip_v5.4.2"
     ),
```
lispandfound: Missing a `/`

```suggestion
genslip_path: Annotated[Path, typer.Option(readable=True, dir_okay=False)] = Path(
        "/EMOD3D/tools/genslip_v5.4.2"
),
```

workflow/scripts/realisation_to_srf.py
```diff
@@ -502,35 +628,42 @@ def generate_srf(
...
+    generic_slip2srf_path: Annotated[
+        Path, typer.Option(readable=True, dir_okay=False)
+    ] = Path("EMOD3D/tools/generic_slip2srf"),
```
lispandfound:
```suggestion
    ] = Path("/EMOD3D/tools/generic_slip2srf"),
```

workflow/realisations.py
```diff
@@ -363,6 +384,36 @@ class SRFConfig(RealisationConfiguration):
...
 
+    point_source_params: PointSourceParams | None
+    """Parameters for point source approximation, if applicable."""
+
+    def __post_init__(self) -> None:
+        """Post-initialization to convert point_source_params dict to PointSourceParams instance."""
+        if self.point_source_params is not None and isinstance(
+            self.point_source_params, dict
+        ):
+            self.point_source_params = PointSourceParams(**self.point_source_params)
```
lispandfound: This should be unnecessary, see schema comment.
```suggestion
```

workflow/realisations.py
```diff
@@ -363,6 +384,36 @@ class SRFConfig(RealisationConfiguration):
     resolution: float
     """The resolution of the SRF geometry"""
 
+    point_source_params: PointSourceParams | None
```
lispandfound: Move `PointSourceParams` class to schema
```suggestion
    point_source_params: schemas.PointSourceParams | None
```

workflow/schemas.py
```diff
@@ -296,6 +335,12 @@ def _corners_to_array(corners_spec: list[dict[str, float]]) -> np.ndarray:
         Literal("resolution", description="Subdivision resolution."): And(
             float, _is_positive
         ),
+        Optional(
+            Literal(
+                "point_source_params",
+                description="Parameters for point source approximation, if applicable",
+            )
+        ): Or(None, POINT_SOURCE_PARAMS_SCHEMA),
```
lispandfound, line 390: After putting point source params class in this module you can now write
```suggestion
        ): Or(None, And(POINT_SOURCE_PARAMS_SCHEMA, Use(PointSourceParams))),
}
```
and remove the `__post_init__` stuff.

workflow/realisations.py
```diff
@@ -348,6 +349,26 @@ def to_dict(self) -> dict[str, Any]:
...
+    risetimefac: float
+    """Rise time factor for generic_slip2srf"""
+
+    risetimedep: float
+    """Rise time depth dependency for generic_slip2srf"""
+
+    inittime: float
+    """Initial time for generic_slip2srf"""
+
+
```
lispandfound:
```suggestion
```

workflow/schemas.py
```diff
@@ -38,6 +40,19 @@
...
+    esg2006 = "esg2006"
+    urs = "urs"
+    ucsb = "ucsb"
+    ucsb2 = "ucsb2"
+    ucsb_T = "ucsb-T"  # noqa: N815
+    ucsb_varT1 = "ucsb-varT1"  # noqa: N815
+    cos = "cos"
+    seki = "seki"
+
+
```
lispandfound, line 56:
```suggestion
@dataclasses.dataclass
class PointSourceParams:
    """Parameters for point source approximation."""

    stype: schemas.Stype
    """Slip type for generic_slip2srf"""

    risetime: float
    """Rise time for generic_slip2srf"""

    risetimefac: float
    """Rise time factor for generic_slip2srf"""

    risetimedep: float
    """Rise time depth dependency for generic_slip2srf"""
    
    inittime: float
    """Initial time for generic_slip2srf"""
```

lispandfound: May need to import `dataclasses`

workflow/scripts/realisation_to_srf.py
```diff
@@ -478,6 +505,10 @@ def generate_fault_srfs_multi(
     environment.srf_directory.mkdir(exist_ok=True)
     environment.gsf_directory.mkdir(exist_ok=True)
     params.velocity_model_1d.write_velocity_model(environment.velocity_model_path)
+
+    # In practice, a Point source will be the only item in `faults`, so if there
+    # is a Point source, restrict to `single_threaded` mode for simplicity.
```
lispandfound: This comment is no longer relevant

```suggestion
```

lispandfound (changes requested):

workflow/scripts/realisation_to_srf.py
```diff
@@ -491,6 +520,101 @@ def generate_fault_srfs_multi(
...
+    None
+        This function does not return a value; it writes the SRF file to disk.
+    """
+
+    if params.srf_config.point_source_params is None:
+        raise ValueError(
+            f"Point source parameters are required for generating SRF for fault '{name}'"
+        )
+
+    environment.gsf_directory.mkdir(parents=True, exist_ok=True)
```
lispandfound: The directory is already created before this point, so this is redundant

workflow/scripts/realisation_to_srf.py
```diff
@@ -109,15 +113,26 @@
...
+        Will create required directories if they do not exist.
     """
     gsf_output_filepath = gsf_output_directory / f"{normalise_name(name)}.gsf"
     gsf_df = gsf.source_to_gsf_dataframe(geometry, subdivision_resolution)
     gsf_df["loc_rake"] = rake
+    if slip is not None:
+        gsf_df["slip"] = slip
+    if init_time is not None:
+        gsf_df["init_time"] = init_time
+    gsf_output_filepath.parent.mkdir(parents=True, exist_ok=True)
```
lispandfound: The directory is already created at this point so this is redundant

workflow/scripts/realisation_to_srf.py
```diff
@@ -109,15 +113,26 @@
...
+        The slip value to write to the GSF file. If None, the slip will not be written.
+    init_time : float, optional
+        The initial time (INIT_TIME) to write to the GSF file.
+        If None, a default value of -1 will be set by `gsf.write_gsf()`.
 
     Returns
     -------
     Path
         The path to the generated GSF file.
+        Will create required directories if they do not exist.
```
lispandfound: The directory for this GSF is already created so this is redundant

lispandfound (changes requested):
Only minor changes and then I am happy to see this merged

lispandfound (dismissed):
LGTM

AndrewRidden-Harper:
@joelridden @lispandfound @sungeunbae This PR has been approved by @lispandfound, but it needs more approvals before it can be merged

workflow/scripts/gcmt_to_realisation.py
```diff
@@ -190,33 +200,83 @@ def gcmt_to_realisation(
...
-        warnings.warn(
-            f"Scaling relationship produced a plane with negative depth ({plane.bounds[:, 2].min()/1000:.2f}km)."
-            " Shifting the plane down to correct."
+    centroid = np.array([latitude, longitude, centroid_depth])
+
+    # Create source based on source_type parameter
+    if source_type == SourceType.POINT_SOURCE:
+        # Create point source
+        magnitude_threshold = 5.0
+        if magnitude < magnitude_threshold:
```
joelridden: Did we do this before for point source? I feel like larger magnitudes have more of the rectangle shape and lower magnitudes being squares. Could be wrong, but I feel like at least I'd ask the question

AndrewRidden-Harper: Thanks @joelridden you raised a great point. Felipe thinks that Brendon just made a typo and meant to write ["Mw>5"](https://uceqeng.slack.com/archives/C03V84ZD21H/p1752041664196469?thread_ts=1747609850.094969&cid=C03V84ZD21H). However, Felipe also thinks we should just use the Leonard rectangular scaling relation for all magnitudes if we have it, so I've made that change. @lispandfound, you might want to check the update

joelridden (approved):
Looks good

workflow/scripts/realisation_to_srf.py
```diff
@@ -504,33 +624,40 @@ def generate_srf(
     genslip_path: Annotated[Path, typer.Option(readable=True, dir_okay=False)] = Path(
         "/EMOD3D/tools/genslip_v5.4.2"
     ),
+    generic_slip2srf_path: Annotated[
+        Path, typer.Option(readable=True, dir_okay=False)
+    ] = Path("/EMOD3D/tools/generic_slip2srf"),
```
sungeunbae, line 629: Is this a valid path?

lispandfound (approved):

workflow/scripts/realisation_to_srf.py
lispandfound: It is in the container it is usually run in.

PR #59, opened 2025-07-22
EMOD3D-compatible Station Coordinates

lispandfound:
Small PR: just ensures that the projection used by the station coordinates stage matches the one expected by EMOD3D by using the new spherical projection code in qcore.


lispandfound:
It also completely removes the `generate-model-coordinates` stage, since it turns out we don't actually need its outputs for any simulations.

AndrewRidden-Harper (approved):

sungeunbae (approved):

joelridden (approved):

PR #65, opened 2025-10-16
Support velocity model version 2.09

lispandfound:
Two changes:

1. Allow support for velocity model version 2.09. This required updating the `nzcvm` code path to account for API changes. When using the `nzcvm` code path, the code now generates the model first in HDF5 format (to allow parallel generation) and then separately converts this to EMOD3D format. The C code remains unchanged and doesn't support 2.09.
2. Relax the floating point requirements for the schema. Previously, a block in a `realisation.json` file must have a floating point value for things like depth (not int). So the following would be invalid:

``` json
{
  "dt": 1
}
```

if the `dt` variable was expected to be a `float`. Now any value that can be coerced to a floating point number via the `float` function is considered valid.


workflow/scripts/generate_velocity_model.py
```diff
@@ -204,10 +224,18 @@ def generate_velocity_model(
...
     else:
         run_nzvm(velocity_model_bin_path, nzvm_config_path, num_threads)
-    shutil.copytree(
-        velocity_model_intermediate_path / "Velocity_Model", velocity_model_output
-    )
+        shutil.copytree(
+            velocity_model_intermediate_path / "Velocity_Model", velocity_model_output
+        )
+
+    shutil.rmtree(velocity_model_intermediate_path)
```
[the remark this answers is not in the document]
lispandfound: `shutil.copytree` *copies* the files so this is totally wrong. But this does make me think, I could move instead of copying.

workflow/scripts/generate_velocity_model.py
```diff
@@ -124,17 +124,42 @@ def run_nzvm(
...
+    num_threads : int | None
+        Number of threads to use (default is inferred by
+        `utils.get_available_cores`)
     """
-    from velocity_modelling.scripts import nzcvm
 
-    nzcvm.generate_velocity_model(nzvm_config_ffp)
+    from velocity_modelling.constants import WriteFormat
+    from velocity_modelling.scripts import generate_3d_model
+    from velocity_modelling.tools import convert_hdf5_to_emod3d
```
joelridden: Do these need to be within the function?

joelridden (commented):
Also need to deal with the ruff import comment

workflow/scripts/generate_velocity_model.py
lispandfound: There was a good reason previously, but I think it no longer applies now.

lispandfound:
Had to fix the CI pipeline to get the tests to pass.

joelridden (approved):

sungeunbae (approved):

PR #69, opened 2025-10-23
Add merge-ts HDF5 output

lispandfound:
The current `merge-ts` stage merges all the xyts files as a big 32-bit floating point array. This is undesirable for a few reasons:

1. No compression. Even though we know that the XYTS files are largely sparse across the domain the naive format includes no mechanism for compression.
2. Full 32-bit precision. This is quite unnecessary for the application of these files (visualisation and sanity checking only). Could a keen-eyed youtube viewer notice the difference between ground motion of  `25.01` cm/s and `25.01005` cm/s? Maybe, but I doubt it. 
3. It saves the x, y, and z components when we only ever visualise magnitude $\sqrt{x^2 + y^2 + z^2}$.
4. Because of the above, the file sizes are huge. Often, this means we don't archive XYTS outputs. Which is a shame because it limits both validation and visualisation of old simulations. 
5. It is yet another "proprietary" file format that we have to maintain infrastructure for. 

This PR addresses the above concerns by adding an option to save xyts.e3d files as a compressed HDF5 file with 16-bit integer quantised outputs. Ground motion in these files is accurate to `0.1` cm/s. Chunk-based compression and shuffling is enabled to further reduce the size of the files. The results can be quite dramatic, here is a dummy simulation I downloaded from RCH with nearly 1GB of xyts outputs (about as small as they get really). 
```bash
~/tmp 
❯ du -h OutBin --exclude='*seis*' 
719M    OutBin
```

Ordinarily, we wouldn't generate an XYTS file for such simulations. But the new code can do this very efficiently

```bash
$ merge-ts hdf5 OutBin test.h5
100%|██████████████████████████████████████████████████████████████████████████████████████| 80/80 [00:00<00:00, 154.38files/s]
$ ls -alh test.h5
.rw-r--r--  4.8M jake 23 Oct 15:35  test.h5
``` 

So we managed to compress 719Mb of data into 4.4Mb of XYTS data, a 100x saving for functionally the same information. Moreover, the files contain far more information. Notably, it uses the spherical projection code to show the lat/lon values of the coordinates.

```python
xr.open_dataset('test.h5')
/home/jake/src/workflow/.venv/lib/python3.13/site-packages/xarray/backends/plugins.py:109: RuntimeWarning: Engine 'gmt' loading failed:
Error loading GMT shared library at 'libgmt.so'.
libgmt.so: cannot open shared object file: No such file or directory
  external_backend_entrypoints = backends_dict_from_pkg(entrypoints_unique)
<xarray.Dataset> Size: 504MB
Dimensions:    (time: 649, y: 265, x: 365)
Coordinates:
  * time       (time) float64 5kB 0.0 0.1 0.2 0.3 0.4 ... 64.5 64.6 64.7 64.8
  * y          (y) int64 2kB 0 1 2 3 4 5 6 7 ... 257 258 259 260 261 262 263 264
  * x          (x) int64 3kB 0 1 2 3 4 5 6 7 ... 357 358 359 360 361 362 363 364
    latitude   (y, x) float64 774kB ...
    longitude  (y, x) float64 774kB ...
Data variables:
    waveform   (time, y, x) float64 502MB ...
Attributes:
    dx:       0.1
    dy:       0.1
    dt:       0.099999994
    mlon:     174.03157
    mlat:     -41.540768
    mrot:     325.42987
```

This HDF5 format will power a new GPU accelerated `plot-ts` to rapidly visualise simulations. There, the int16 representation also means that the memory efficiency of this code is considerably faster.

### DIFF
--- a/workflow/scripts/merge_ts.py
+++ b/workflow/scripts/merge_ts.py
@@ -30,14 +30,16 @@ import os
 from pathlib import Path
 from typing import Annotated
 
-import typer
+import numpy as np
 import tqdm
+import typer
 import xarray as xr
 
-from qcore import cli, xyts, coordinates
+from qcore import cli, coordinates, xyts
 from workflow.scripts import merge_ts_loop
 
 app = typer.Typer()
+
 
 def merge_ts_xyts(
     component_xyts_directory: Annotated[
@@ -54,7 +56,7 @@ def merge_ts_xyts(
         typer.Argument(dir_okay=False, writable=True),
     ],
     glob_pattern: str = "*xyts-*.e3d",
-):
+) -> None:
     """Merge XYTS files.
 
     Parameters
@@ -153,8 +155,8 @@ def merge_ts_hdf5(
         typer.Argument(dir_okay=False, writable=True),
     ],
     glob_pattern: str = "*xyts-*.e3d",
-    complevel: int = 4
-):
+    complevel: int = 4,
+) -> None:
     """Merge XYTS files.
 
     Parameters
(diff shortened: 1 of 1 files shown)


AndrewRidden-Harper (approved):
very nice

joelridden (approved):

PR #75, opened 2025-12-01
Unify resolution and dt

lispandfound:
This PR removes the duplicated spatial and temporal resolution calculations to avoid typos like the default `genslip_dt` argument.

The `dt` argument is always set as a function of spatial resolution, and is defined as `resolution / 20`. The `nx`, `ny`, `nz` are moved into `create_e3d_par` stage.


workflow/default_parameters/develop/defaults.yaml
lispandfound, line 1: **NOTE**: The develops defaults have updated. Nobody actually uses these at the moment. Don't bother reviewing these.

AndrewRidden-Harper (dismissed):

joelridden (dismissed):

sungeunbae:
This is a good change - just fix the type check.
Screenshot 2025-12-02 at 8 37 01 PM

lispandfound:
@sungeunbae type checking is fixed in the other PR, don't want to copy the work over

lispandfound:
Just one tiny HF config merge conflict.

joelridden (approved):

sungeunbae (approved):

PR #77, opened 2025-12-11
Fix package references

lispandfound:
Just swaps to PyPi references where we have packages


lispandfound:
Closed in favour of #78

PR #79, opened 2025-12-16
Remove merge_ts

claudio525:


workflow/scripts/plan_workflow.py
sungeunbae: This line should go

sungeunbae: this line should go

sungeunbae: not needed?

sungeunbae (changes requested):

pyproject.toml
lispandfound, line 1: You also need to remove the `cython` requirement at the top of the `pyproject.toml` file. This is added to build `merge-ts` but is no longer needed.

lispandfound (changes requested):
Also delete the `templates` folder in the workflow package.

sungeunbae (approved):

lispandfound (approved):

PR #80, opened 2025-12-16
Rupture velocity refactor

lispandfound:
This PR refactors `hf-sim` and `realisation-to-srf` to have a common set of rupture velocity parameters. The background is explained [in slack](https://uceqeng.slack.com/archives/C09MJ5ECCF8/p1765772316295049).

Rupture velocity determines the how the wavefront of the rupture moves across the rupture geometry during SRF generation, and HF calculation. It relates to the SRF rupture initiation times, and the lower frequency HF output. In the current workflow settings the rupture velocity is decreased near the surface and at depth. It is consistently adjusted in the high-frequency code path too. However, this is only because Rob applied a consistent set of defaults between the two codebases. When we adjust these parameters we need to make sure they stay in sync. This is the motivation for the PR.

I also slightly refactored the hf-sim code to add some tests. Basically cutting down the main function to have it call more subfunctions and then testing those subfunctions. 

**The high line count here is just because I have extracted common options into a new defaults file.** I did this because I got sick of editing four different files in the same places over and over.


workflow/scripts/hf_sim.py
```diff
@@ -236,6 +272,110 @@ def stable_hash(station: str) -> int:
...
+        not matter. The station seeds are based on their name only.
+
+    Returns
+    -------
+    npt.NDArray[np.int32]
+        A list of station seeds.
+    """
+    station_hashes = np.array([stable_hash(name) for name in stations], dtype=np.int32)
+    # Rather than add (which could overflow and cause annoying numpy
+    # warnings), we just xor the hf seed with the station hashes.
```
AndrewRidden-Harper, line 293: nice idea!

AndrewRidden-Harper (dismissed):

joelridden (dismissed):

joelridden (approved):

AndrewRidden-Harper (approved):

PR #81, opened 2025-12-17
Expose `generate_velocity_model_parameters` so that you do not have to write to realisation

lispandfound:
Simple change. Currently the only way to generate a domain is first write a realisation file and then execute the generate velocity model parameters script. This small refactor changes this so that domain generation can be invoked with simple realisation objects and without JSON serialisation.

### DIFF
--- a/workflow/scripts/generate_velocity_model_parameters.py
+++ b/workflow/scripts/generate_velocity_model_parameters.py
@@ -426,9 +426,117 @@ def pgv_target(
     )
 
 
+def generate_velocity_model_parameters(
+    source_config: SourceConfig,
+    velocity_model_parameters: VelocityModelParameters,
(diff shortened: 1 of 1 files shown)


workflow/scripts/generate_velocity_model_parameters.py
```diff
@@ -426,45 +426,57 @@ def pgv_target(
     )
 
 
-@cli.from_docstring(app)
-@log_utils.log_call()
 def generate_velocity_model_parameters(
```
claudio525, line 429: I think it would be a good idea to put this into a module instead of leaving it as part of the script?

lispandfound:
Closed in favour of #83 which also includes the rrup simplification.

PR #83, opened 2026-01-06
Refactor generate velocity model parameters

lispandfound:
1. It removes the `dict_zip` function and moves it to utils. Doing this allowed proper testing of the function. I also added better type hinting in the two and three argument case.
2. Fixed the rake calculation to use a moment-weighted circular mean instead of unweighted arithmetic mean for GMM inputs.
3. Depth calculation is now always at least as deep as the deepest source plus a buffer.
4. R-surface is used instead of rrup, and is estimated directly from magnitude instead of PGV. This reflects changes in our thinking on domain estimation in the last year or so.

I also added some unit and integration tests to the generate_domain module, which test some key properties:

1. That the simulation domain contains the sources,
2. That the simulation duration scales with magnitude and produces sensible outputs and,
3. That a simple E2E test of the generate domain stage works.

Finally, in response to @claudio525, I have extracted the domain generation logic into a function that does not require a realisation file. There is now `generate_domain.generate_domain` which takes realisation *objects* and returns domain parameters, and *generate_domain.generate_domain_from_realisation* that takes a path to a realisation file and reads the relevant parameters from the file, and writes them back.


pyproject.toml
```diff
@@ -17,38 +17,38 @@ dependencies = [
...
   "schema",    # For loading realisations
   "structlog", # Logging.
   "psutil",    # To get the CPU affinity for jobs
+
 ]
 
 [project.optional-dependencies]
-test = ["pytest"]
+test = [
+  "pytest>=6.0.0",            # required for the tool.pytest section
```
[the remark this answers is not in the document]
lispandfound: This is wrong.

pyproject.toml
```diff
@@ -84,6 +84,8 @@ workflow = "workflow"
 
 [tool.setuptools_scm]
 
+[tool.pytest]
```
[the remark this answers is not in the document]
lispandfound: nope

workflow/scripts/generate_domain.py
```diff
@@ -0,0 +1,590 @@
...
+        Bottom depth (in km).
+
+    Returns
+    -------
+    float
+        The maximum simulation depth.
+
+    References
+    ----------
+    See the "Custom Models Used in VM Params" wiki page for an explanation of this function.
```
sungeunbae: Where is this wiki?

workflow/scripts/generate_domain.py
```diff
@@ -0,0 +1,590 @@
...
+    Returns
+    -------
+    BoundingBox
+        The smallest domain containing the sources and areas within
+        the estimated surface distance for the given magnitudes.
+    """
+    # This polygon includes all the faults corners + a 2km buffer (which must be in the simulation domain).
+
+    fault_buffer_polygons = [
+        shapely.buffer(fault.geometry, 2000)
```
sungeunbae: Is there any reason why 2km buffer is good?

AndrewRidden-Harper (dismissed):

workflow/scripts/generate_domain.py
lispandfound: 2km buffer is enough to separate the source from the boundary points of the domain which according to Brendon are treated differently in the FD simulation and so the source should not be contained in them.

workflow/scripts/generate_domain.py
lispandfound: Great question! I can't find that page anymore. I've updated the doc with more detailed  and linked to a slack message. The implementation has already drifted from that message but I include it for historicity.

workflow/scripts/generate_domain.py
lispandfound: I've added it as a parameter now.

AndrewRidden-Harper (dismissed):

workflow/scripts/generate_domain.py
```diff
@@ -0,0 +1,609 @@
...
+        Buffer width (km) around sources in rupture. Domain edge is
+        guaranteed not be within this distance from any source.
+
+    Returns
+    -------
+    BoundingBox
+        The smallest domain containing the sources and areas within
+        the estimated surface distance for the given magnitudes.
+    """
+    # This polygon includes all the faults corners + a 2km buffer (which must be in the simulation domain).
```
sungeunbae: Could fault_buffer have a default 2, and edit the comment "all the faults corners + a 2km (default) buffer"

lispandfound: Fault buffer has a default of two already. I can update the comment.

AndrewRidden-Harper (approved):

sungeunbae (approved):

workflow/scripts/generate_domain.py
```diff
@@ -0,0 +1,610 @@
...
+------
+A realisation file containing a metadata configuration, source definitions and rupture propagation information.
+
+Outputs
+-------
+A realisation file containing velocity model and domain extent parameters.
+
+Environment
+-----------
+Can be run in the cybershake container. Can also be run from your own computer using the `generate-domain` command which is installed after running `pip install workflow@git+https://github.com/ucgmsim/workflow`.
```
claudio525, line 18: Nitpick, but maybe this should use pypi package instead of pulling it from github? Would need to do a release after this commit, but seems like the nicer option.

claudio525 (commented):
Nice one, just one minor comment.

workflow/scripts/generate_domain.py
lispandfound: At some point yes I will do this.

PR #84, opened 2026-01-06
Duplicate the last layer of the HF 1D VM to stop crashes with subduction runs

lispandfound:
Pretty simple, noticed by @claudio525 first in his subduction runs. If the HF raycasting requires travels through the last layer then the code will crash. The simplest fix that doesn't require changing the HF fortran code is just duplicating the last layer. I believe I saw this when porting from the old workflow and assumed it was a typo... It doesn't have any impact on simulations that do not propagate to the last layer (i.e. most of them)


AndrewRidden-Harper (approved):

sungeunbae (approved):
Looks like the best solution for now.

PR #94, opened 2026-03-02
fix: dropping eas component in IM calculation

lispandfound:
We are currently dropping the EAS component of im calculation, this small PR fixes that by adding eas to the initial dataset calculation


joelridden (approved):
Maybe look into the test as to why the hf input code is failing though

AndrewRidden-Harper (approved):

PR #105, opened 2026-06-03
Use BoldM type for fault magnitudes in Magnitudes config

AndrewRidden-Harper:
## Summary

`source_modelling` 2026.6.1 introduces the `BoldM` magnitude `NewType`. This
updates the `Magnitudes` realisation config to use `BoldM` instead of `float`,
and makes call sites across the codebase consistent.

## Changes
- Bump `source_modelling` dependency to `>=2026.6.1` (`pyproject.toml`, `uv.lock`)
- `Magnitudes.magnitudes` and `__getitem__` now use `BoldM`
- Cast floats to `BoldM` when constructing `Magnitudes` (`gcmt_to_realisation`,
  `nshm2022_to_realisation`, tests)
- Pass `bold_m=True` to `moment.magnitude_to_moment` for `BoldM` magnitudes,
  preserving the existing runtime default (`generate_domain`, `realisation_to_srf`)

## Notes
- In `nshm2022_to_realisation`, the magnitudes dict is shared with
  `nshmdb.most_likely_fault` (whose `dict[str, float]` is invariant), so the
  `Magnitudes(...)` construction uses a scoped `# ty: ignore[invalid-argument-type]`
  — safe because `BoldM` is a `float` `NewType` (identity at runtime).

### DIFF
--- a/workflow/scripts/gcmt_to_realisation.py
+++ b/workflow/scripts/gcmt_to_realisation.py
@@ -284,7 +284,9 @@ def gcmt_to_realisation(
         hypocentre = np.array([1 / 2, 1 / 2])
 
     source_config = SourceConfig(source_geometries={gcmt_event_id: source_geometry})
-    magnitudes = Magnitudes(magnitudes={gcmt_event_id: float(magnitude)})
+    magnitudes = Magnitudes(
+        magnitudes={gcmt_event_id: magnitude_scaling.BoldM(float(magnitude))}
+    )
     rakes = Rakes(rakes={gcmt_event_id: float(rake)})
     rupture_config = RupturePropagationConfig(
         rupture_causality_tree={gcmt_event_id: None},
--- a/workflow/scripts/gcmt_to_realisation.py
+++ b/workflow/scripts/gcmt_to_realisation.py
@@ -192,7 +192,7 @@ def gcmt_to_realisation(
             param_hint="GCMT_EVENT_ID",
         )
 
-    magnitude = moment.moment_to_magnitude(solution_moment)
+    magnitude = moment.moment_to_magnitude(solution_moment, bold_m=True)
 
     model = community_fault_model.get_community_fault_model()
 
--- a/workflow/scripts/gcmt_to_realisation.py
+++ b/workflow/scripts/gcmt_to_realisation.py
@@ -285,7 +285,7 @@ def gcmt_to_realisation(
 
     source_config = SourceConfig(source_geometries={gcmt_event_id: source_geometry})
     magnitudes = Magnitudes(
-        magnitudes={gcmt_event_id: magnitude_scaling.BoldM(float(magnitude))}
+        magnitudes={gcmt_event_id: magnitude}
     )
     rakes = Rakes(rakes={gcmt_event_id: float(rake)})
     rupture_config = RupturePropagationConfig(
--- a/workflow/scripts/gcmt_to_realisation.py
+++ b/workflow/scripts/gcmt_to_realisation.py
@@ -192,7 +192,9 @@ def gcmt_to_realisation(
             param_hint="GCMT_EVENT_ID",
         )
 
-    magnitude = moment.moment_to_magnitude(solution_moment, bold_m=True)
+    magnitude = moment.moment_to_magnitude(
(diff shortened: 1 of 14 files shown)


workflow/scripts/gcmt_to_realisation.py
```diff
@@ -284,7 +284,9 @@ def gcmt_to_realisation(
         hypocentre = np.array([1 / 2, 1 / 2])
 
     source_config = SourceConfig(source_geometries={gcmt_event_id: source_geometry})
-    magnitudes = Magnitudes(magnitudes={gcmt_event_id: float(magnitude)})
+    magnitudes = Magnitudes(
+        magnitudes={gcmt_event_id: magnitude_scaling.BoldM(float(magnitude))}
+    )
```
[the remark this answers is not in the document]
AndrewRidden-Harper: Thanks, but this is a false positive — there's no `NameError` here. `magnitude_scaling` is already imported at the top of the file (line 43: `from source_modelling import community_fault_model, magnitude_scaling, moment, sources`) and is used elsewhere (e.g. lines 112, 214). This diff only added `magnitude_scaling.BoldM(...)`, reusing the existing module import. CI (tests + typecheck) passes, confirming the name resolves. Leaving as-is.

workflow/scripts/nshm2022_to_realisation.py
```diff
@@ -376,7 +376,9 @@ def generate_realisation(
         strategy=str(strategy),  # type: ignore
         jump_impossibility_limit_distance=round(jump_cutoff * 1000),
     )
-    magnitudes = Magnitudes(magnitudes)
+    # ty flags this because dict is invariant (dict[str, float] != dict[str, BoldM]),
+    # but BoldM is a float NewType, so this is a safe no-op at runtime.
+    magnitudes = Magnitudes(magnitudes)  # ty: ignore[invalid-argument-type]
```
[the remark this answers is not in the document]
AndrewRidden-Harper: This project's type checker is Astral's [`ty`](https://github.com/astral-sh/ty), not mypy/pyright — CI runs `uv run ty check` (see `.github/workflows/types.yml`). So `# ty: ignore[invalid-argument-type]` is the correct directive: `ty` is the tool name (not a typo for `type`), and `invalid-argument-type` is ty's actual rule name (`arg-type` is mypy's code, which ty doesn't use). The `typecheck` CI job passes with this exact syntax. Switching to `# type: ignore[arg-type]` would use a code ty doesn't recognize and make the suppression less precise, so keeping the ty-specific form.

workflow/scripts/gcmt_to_realisation.py
```diff
@@ -284,7 +284,9 @@ def gcmt_to_realisation(
         hypocentre = np.array([1 / 2, 1 / 2])
 
     source_config = SourceConfig(source_geometries={gcmt_event_id: source_geometry})
-    magnitudes = Magnitudes(magnitudes={gcmt_event_id: float(magnitude)})
+    magnitudes = Magnitudes(
+        magnitudes={gcmt_event_id: magnitude_scaling.BoldM(float(magnitude))}
```
lispandfound:
```suggestion
        magnitudes={gcmt_event_id: magnitude}
```

workflow/scripts/gcmt_to_realisation.py
lispandfound:
```suggestion
    magnitude = moment.moment_to_magnitude(solution_moment, bold_m=True)
```
May help to add this, just to make it explicit.

lispandfound (changes requested):
LGTM barring the conversion of magnitude types in `gcmt-to-realisation`.

workflow/scripts/nshm2022_to_realisation.py
lispandfound: This conversion is catching a bug in the current implementation! The current code is using `qcore` magnitude scaling relations. These are problematic because they don't contain our fixes to deal with the different equations which live in source modelling. We need to:

1. Change the magnitude scaling relationships to use source modelling in is the module,
2. Remove the legacy leonard mw to a relationship and use the one in source modelling. 

This should overall reduce the line count and remove the need for this type index.

lispandfound (dismissed):

lispandfound (approved):

sungeunbae (approved):

PR #114, opened 2026-06-23
SW4 station post-processing

lispandfound:
Adds a function to read the station HDF5 output from SW4 and convert it into a compliant NetCDF file. This file is compatible with all downstream processing in hf-sim, bb-sim, so this is the only portion of the workflow required to drop-in SW4 instead of EMOD3D (besides the velocity model).


lispandfound:
@gemini-code-assist please re-review

joelridden (dismissed):

AndrewRidden-Harper (dismissed):

workflow/scripts/lf_to_xarray.py
```diff
@@ -26,30 +26,127 @@
...
+
+    dt = np.float32(handle["DELTA"][()])
+    xs = []
+    ys = []
+    zs = []
+    stations = []
+    attributes = {"start_sec": 0.0, "dt": dt}
+    for station_name, group in handle.items():
+        if "NPTS" not in group:
+            continue
```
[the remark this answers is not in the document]
sungeunbae: This looks like a valid comment from gemini?

workflow/scripts/bb_sim.py
```diff
@@ -113,6 +114,51 @@ def align_waveforms(
...
+        Resampled dataset with updated time coordinates and dt attribute.
+    """
+    duration = dset["waveform"].sizes["time"] * dset.attrs["dt"]
+    nt = round(duration / dt)
+
+    # NOTE: I am not providing a default start second because we consider it an
+    # error not to provide one (no implicit magic behaviour).
+    new_time = np.arange(nt) * dt + dset.attrs["start_sec"]
+
+    resampled_waveform = xr.apply_ufunc(
```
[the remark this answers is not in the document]
sungeunbae: @lispandfound  Does this make sense? I am not sure

workflow/scripts/lf_to_xarray.py
```diff
@@ -26,30 +26,127 @@
...
+            raise RuntimeError(f"SW4 output is corrupted: {npts=} but {global_npts=}")
+        global_npts = npts
+        # Dask arrays here ensure that data is read chunkwise from the HDF5 file
+        # without putting it all in-memory.
+        x = da.from_array(group["X"], chunks=-1)
+        xs.append(x * CMS)
+        y = da.from_array(group["Y"], chunks=-1)
+        ys.append(y * CMS)
+        z = da.from_array(group["Z"], chunks=-1)
+        zs.append(z * CMS)
```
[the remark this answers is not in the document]
sungeunbae: @lispandfound  Please comment why you thought it doesn't matter?

sungeunbae (commented):
Some explanation will be desirable when you dismiss comments

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
+    with h5py.File(sw4_ffp, "r") as handle:
+        for i, station_name in enumerate(stations):
+            group = handle[station_name.item()]
+            if "NS" not in group:
+                raise RuntimeError(
+                    f"Station {station_name.item()} has no EW/NS/UP datasets."
+                    " The SW4 rechdf5 command must output geographic (NSEW)"
+                    " displacement-mode components."
+                )
+            waveforms[0, i] = group["EW"][:]
```
lispandfound: The components are stored as x=EW, y=NS, z=UP. The rest of the pipeline expects x=north, y=east, z=down: qcore `_postprocess_waveform` says "x points north, y points east, and z points down", and the IM package labels component 0 as `000` and component 1 as `090`.

**Failure scenario:** Convert an SW4 recording and run im-calc. `PGA_000`, `pSA_000` and the other `000` IMs are computed from the east-west trace, and the `090` IMs from north-south, so they don't match EMOD3D runs of the same event. bb-sim pairs LF and HF by component index, so it would also add SW4 east-west LF to HF north-south, and SW4 up to a z axis that points down. Suggested fix: `waveforms[0] = NS`, `waveforms[1] = EW`, `waveforms[2] = -UP`.

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
+
+    time = np.arange(global_npts) * dt
+    return xr.Dataset(
+        {
+            "lat": ("station", latitudes),
+            "lon": ("station", longitudes),
+        },
+        coords={
+            "station": stations,
+            "component": ["x", "y", "z"],
```
lispandfound: The SW4 dataset has no per-station `x`/`y` coordinates, but bb-sim reads `lf.x.values` and `lf.y.values`. bb-sim also rebuilds its output without `supergrid_depth(_gp)` or the supergrid width attributes.

**Failure scenario:** Running `lf-to-xarray --format sw4` and then bb-sim raises `AttributeError: 'Dataset' object has no attribute 'x'` in the output coords block of bb_sim.py. Even if that is fixed, the BB file drops the supergrid data, so im-calc on broadband output only gets the all-NaN fallback. The supergrid feature only works when im-calc is run on the raw LF file.

lispandfound (commented):
Code review by Claude Code (run locally, since the review action is currently broken). 9 finding(s).

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
+    """Build the dataset's coordinates and attributes from an SW4 station file."""
+    global_npts = None
+    stations = []
+    latitudes = []
+    longitudes = []
+    supergrid_depths = []
+    supergrid_depths_gp = []
+
+    with h5py.File(sw4_ffp, "r") as handle:
+        dt = np.float32(handle["DELTA"][:].squeeze())
```
lispandfound: (Unverified) This assumes `DELTA`, `NPTS`, `STLA,STLO,STDP`, `SGDEPTH` and `SGWIDTH` are 1-element datasets read with `[:]`. SW4's HDF5 station output may store some of them as HDF5 attributes. Please check against a real SW4 file.

**Failure scenario:** If they are attributes, `handle["DELTA"]` raises `KeyError`, or `"NPTS" not in group` skips every station and the run fails with 'No valid station recordings found'. If any is a scalar dataset, `[:]` raises 'Illegal slicing argument for scalar dataspace'. The test fixture only covers the 1-element dataset layout. Reading with `[()]`, falling back to `.attrs`, would handle every case.

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
+
+        attrs: dict[str, np.float32 | float] = {"dt": dt}
+        for sw4_name, attribute_name in SUPERGRID_WIDTH_ATTRIBUTES.items():
+            if sw4_name in handle:
+                attrs[attribute_name] = float(handle[sw4_name][:].squeeze())
+        for station_name, group in handle.items():
+            if "NPTS" not in group:
```
lispandfound: `handle.items()` returns root datasets as well as station groups. On an h5py `Dataset`, `"NPTS" not in group` falls back to iterating the data.

**Failure scenario:** A scalar root dataset, such as a scalar `SGWIDTH` or `DATETIME`, raises `TypeError: Can't iterate over a scalar dataset`; this reproduces. A large root dataset gets read in full and compared element by element. Check `isinstance(group, h5py.Group)` first.

workflow/scripts/im_calc.py
```diff
@@ -764,6 +807,8 @@ def calculate_intensity_measures(
             station=broadband.station.str.match(r"^(\w{4})$").values
         )
 
+    supergrid = _supergrid_coordinates(broadband)
```
lispandfound: `_supergrid_coordinates` returns all-NaN `supergrid_depth` and `supergrid_depth_gp` arrays when the input doesn't have them, so every EMOD3D IM file now carries two NaN coordinates.

**Failure scenario:** Every existing EMOD3D run gains `supergrid_depth=NaN` on each IM dataset. Downstream concatenation or merging with older IM outputs sees extra coordinates, and NaN coordinates break `equals`/`identical` comparisons. Returning `{}` when the coordinates are absent avoids this and removes the need for `absent()`.

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
+        chunked_stations,
+        kwargs={
+            "time": dset["time"],
+            "component": dset["component"],
+            "sw4_ffp": sw4_ffp,
+        },
+        template=_template_waveform(dset, batch_size),
+    )
+
+    waveform = (waveform * CMS).differentiate("time")
```
lispandfound: `TARGET_CHUNK_BYTES` sets the size of each chunk but not total memory. Each 128 MiB batch also gets a `* CMS` copy and a `differentiate` output, and the threaded scheduler runs one batch per core.

**Failure scenario:** On a 64-core node, peak memory can reach several 128 MiB arrays per task times 64 tasks, which is tens of GiB. Doing the scaling and `np.gradient` inside `_read_station_batch` in place, or limiting the dask worker count, keeps memory bounded.

workflow/scripts/im_calc.py
```diff
@@ -161,6 +161,49 @@
 }
 
 
+SUPERGRID_COORDINATES = ("supergrid_depth", "supergrid_depth_gp")
+"""Records the SW4 depth inside the supergrid (in metres and gridpoints)."""
+
+SUPERGRID_WIDTH_ATTRIBUTES = ("supergrid_width", "supergrid_width_gp")
```
lispandfound: `SUPERGRID_WIDTH_ATTRIBUTES` is defined twice with the names hard-coded: as a tuple here and as a dict in `lf_to_xarray`.

**Failure scenario:** These names have already changed once in this PR, in 'Name the supergrid width attributes'. If one side is renamed again, im-calc silently stops copying the widths, and no test covers both sides. Deriving it with `tuple(lf_to_xarray.SUPERGRID_WIDTH_ATTRIBUTES.values())`, or sharing a single constant, removes the duplication.

workflow/scripts/lf_to_xarray.py
```diff
@@ -25,30 +25,230 @@
...
 
     Parameters
     ----------
-    lfseis_directory : Path
+    low_frequency_path : Path
         Directory containing station seismogram outputs.
     output_ffp : Path
         Path to write the xarray dataset
+    format : Format, optional
+        Format for the low-frequency inputs (EMOD3D or SW4). If format is SW4,
```
lispandfound: The docs contradict each other: `low_frequency_path` is still described as 'Directory containing station seismogram outputs', and the module usage line still says `OUTBIN_DIRECTORY`. For SW4 this argument must be a file.

**Failure scenario:** A user who follows `--help` passes a directory with `--format sw4` and gets 'SW4 format requires station recording file.'

tests/test_lf_to_xarray.py
```diff
@@ -0,0 +1,145 @@
...
+    """Run `lf-to-xarray` on an SW4 station file and read the result back."""
+    output = sw4_ffp.with_suffix(".nc")
+    lf_to_xarray.convert_lf_to_xarray_dataset(
+        sw4_ffp, output, format=lf_to_xarray.Format.SW4
+    )
+    with xr.open_dataset(output, mask_and_scale=False) as dataset:
+        return dataset.load()
+
+
+def test_supergrid_penetration_arrives_as_float32_coordinates(tmp_path: Path) -> None:
```
lispandfound: The tests only check supergrid coordinates and attributes. Nothing checks waveform values, component order, units, or the im-calc supergrid path.

**Failure scenario:** The component swap, a missing `* CMS`, or a broken derivative would all pass. The im-calc path, `_supergrid_attributes` with `absorbing_layer` and the width propagation, has no test.

workflow/scripts/lf_to_xarray.py
lispandfound: I have checked these and it works correctly.

workflow/scripts/lf_to_xarray.py
lispandfound: In this case the docstring is wrong in qcore.

tests/test_lf_to_xarray.py
lispandfound: This is fine.

workflow/scripts/lf_to_xarray.py
lispandfound: This is addressed in a PR further up the stack.

workflow/scripts/im_calc.py
lispandfound: Fixed. I have moved these attributes to the sw4 module.

AndrewRidden-Harper (approved):
look great

PR #121, opened 2026-07-13
Multi-segment SRF version fix

lispandfound:
Multi-segment SRFs now use the version of the SRF that matches the fault-segment SRFs that generated them, and throws an error if the SRF versions don't match.


AndrewRidden-Harper (approved):
perfect

joelridden (approved):

PR #124, opened 2026-07-16
Bump the source_modelling version in Docker container image to get the SRF fixes

AndrewRidden-Harper:
This is not strictly needed because the previous `source_modelling>=2026.07.1` automatically installed the latest `2026.07.2`, but I'm just doing the update for completeness and consistency

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -74,8 +74,10 @@ ENV LD_LIBRARY_PATH="/usr/local/lib:$LD_LIBRARY_PATH"
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
 # Setup Virtual Environment and Install via pip
+# openquake.engine 3.26+ requires gdal~=3.13.1, which is sdist-only and refuses to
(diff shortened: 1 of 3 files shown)


AndrewRidden-Harper:
### Why there's a Dockerfile commit in here

Adding an unrelated fix for the container build, which has been broken since 2026-07-13.

`openquake.engine` 3.26.0, released 2026-07-13, is the first release to pin GDAL (`gdal~=3.13.1`). `gdal` ships sdist-only, so pip compiles it against the system libgdal and hard-errors:

```
Python bindings of GDAL 3.13.1 require at least libgdal 3.13.1, but 3.12.2 was found
```

Our Ubuntu base has libgdal 3.12.2. Nothing in this repo changed — the image resolves its Python dependencies fresh from PyPI on every build, so one upstream release was enough to break it.

The fix caps `openquake.engine<3.26`. That's safe: 3.25.1 declares no gdal dependency, `oq_wrapper` only needs `>=3.21`, and 3.25.1 is what the last successful image already shipped — so it holds the current line rather than downgrading anything.

lispandfound (approved):

PR #127, opened 2026-07-21
Add empirical calculations to intensity measures

lispandfound:
Adds empirical calculations to intensity measure outputs where empirical models support a given tectonic type and intensity measure. 

To support this structure, I have also refactored im calc so that it finally outputs in xarray data tree format instead of dataset format. The upshot of doing this is that we can uncouple the components so that each intensity measure carries only components it actually computes. Especially for models like NSHM2022 that only support pSA it's kind of silly to have every intensity measure carry all-NaN EAS and empirical model components. The Vs30 is carried through from the broadband calculations. All input parameters (distance metrics, average rake, average dip, etc etc) are injected into the datatree output for completeness. The defaults target a tectonic type of active shallow and NSHM2022. This is not tectonic type aware and I will address this shortly when the subduction defaults are finalised.


AndrewRidden-Harper (approved):

workflow/scripts/im_calc.py
```diff
@@ -111,6 +279,423 @@ def _trace_polygon(source_geometries: dict[str, IsSource]) -> shapely.Geometry:
...
+    source_config: SourceConfig,
+    magnitudes: Magnitudes,
+    rakes: Rakes,
+    hypocentre: np.ndarray,
+) -> SourceParameters:
+    """Reduce a multi-fault realisation to a single set of rupture parameters.
+
+    Ground motion models describe a rupture with a single magnitude, rake,
+    dip and depth. Multi-fault realisations are collapsed into these by
+    averaging each fault's contribution, weighted by its moment.
```
joelridden, line 437: Based on what we have done before for empirical results for multi fault events we averaged based on the area of the fault not the moment. There was a few discussions regarding what to use and we landed on using area, however could be open to reconsider this adjustment if there has been a new development for this.

https://github.com/ucgmsim/Empirical_Engine/blob/1d39ae9d620767969f71b63748bbac9eea4e18d6/oq_wrapper/estimations.py#L67

workflow/scripts/im_calc.py
```diff
@@ -111,6 +279,423 @@ def _trace_polygon(source_geometries: dict[str, IsSource]) -> shapely.Geometry:
...
+    vs30 : xr.DataArray
+        The per-station vs30 values (m/s).
+
+    Returns
+    -------
+    SiteParameters
+        The site parameters, with basin depths estimated using the Chiou
+        and Youngs (2008) relations.
+    """
+    z1pt0 = chiou_young_08_calc_z1p0(vs30)  # ty: ignore[invalid-argument-type]
```
joelridden, line 535: Do we want to estimate for all of the sites? Wondering if we have any "real sites" in this where we have a better measured / estimate value of z1.0 / 2.5 etc to use those instead from the site database.

lispandfound: Yeah this is a stand-in at the moment, I intend to replace this with actual site database values at a later date.

workflow/scripts/im_calc.py
lispandfound: Ah didn't see this, I will change the calculation to use that averaging.

lispandfound:
Folded into #139 (the empirical commit now sits at the tip of `nzvm/im-calc-metadata`); #145 is retargeted onto `nzvm/bb-matched-filter`.

PR #128, opened 2026-07-29
Port srf2stoch to Python

lispandfound:
Blah blah summary will update

### DIFF
--- a/workflow/scripts/generate_stoch.py
+++ b/workflow/scripts/generate_stoch.py
@@ -18,7 +18,7 @@ Usage
 
 Environment
 -----------
-Can be run in the cybershake container. Can also be run from your own computer using the `generate-stoch` command which is installed after running `pip install workflow@git+https://github.com/ucgmsim/workflow`. If you are executing on your own computer you also need to specify the `srf2stoch` path (`--srf2stoch-path`).
+Can be run in the cybershake container. Can also be run from your own computer using the `generate-stoch` command which is installed after running `pip install workflow@git+https://github.com/ucgmsim/workflow`.
 
 For More Help
 -------------
@@ -73,10 +73,14 @@ def _box_average_matrix(
     -----
     The weights correspond exactly to the fractional overlap of coarse bins over
     fine bins. This is mathematically equivalent to upsampling both grids to
-    their Least Common Multiple (LCM) base units and computing a standard block
-    average (which is the approach that srf2stoch.c takes). The main advantage
-    of this approach is that a sparse matrix does not have to materialise all
-    the empty cell overlaps in memory.
+    their Least Common Multiple (LCM) base units, padding the geometry with
+    zeros in these units and computing a standard block average. The special
+    case where the fine grid and coarse grid span the same length is implemented
+    by srf2stoch.c. The main advantage of this approach is that a sparse matrix
+    does not have to materialise all the empty cell overlaps in memory. A
+    secondary advantage is that we handle the padded case, which lets us set a
+    uniform dx/dy for all SRF segments as the HF code demands without changing
+    total moment.
 
     For example, downsampling 5 fine cells to 3 coarse cells implies an LCM of
     15 base units. The 5 fine cells (A-E) take up 3 units each, while the 3
@@ -110,7 +114,7 @@ def _box_average_matrix(
       units of D (2/5) and all 3 units of E (3/5).
 
     """
-    bin_width = fine_dx / coarse_dx
+    bin_width = coarse_dx / fine_dx
     edges = np.arange(n_coarse + 1) * bin_width
     rows, cols, weights = [], [], []
     for j in range(n_coarse):
@@ -135,21 +139,41 @@ def circular_mean(angles: np.ndarray, weights: np.ndarray) -> float:
     weights : array of floats
         The weights to apply to the average.
 
-
     Returns
     -------
     float
         The weighted circular mean of angles.
     """
 
-    rad = np.radians(angles)
+    rad = np.radians(np.ravel(angles))
     x = np.cos(rad)
     y = np.sin(rad)
+    weights = np.ravel(weights)
+    if not weights.any():
+        # A plane with no slip anywhere has no slip-weighted mean, so
+        # fall back to an unweighted average of the angles.
+        weights = np.ones_like(weights)
     avg_vector = np.average(np.c_[x, y], weights=weights, axis=0)
-    return np.arctan2(avg_vector[1], avg_vector[0]).item()
+    return np.degrees(np.arctan2(avg_vector[1], avg_vector[0])).item() % 360.0
 
 
 def convert_srf_to_stoch(srf_file: SrfFile, dx: float, dy: float) -> StochFile:
+    """Convert an SRF file into a Stoch file by box-averaging slip, tinit and tinit * rise.
+
+    Parameters
+    ----------
+    srf_file : SrfFile
+        The SRF file to convert.
+    dx : float
+        The desired strike-resolution for the output stoch file.
+    dy : float
+        The desired dip-resolution for the output stoch file.
+
+    Returns
+    -------
+    StochFile
+        An output stoch file downsampled from ``srf_file``.
+    """
     planes = []
     for i, segment in enumerate(srf_file.segments):
         header = srf_file.header.iloc[i].astype(np.float32)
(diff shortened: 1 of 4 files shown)


workflow/scripts/generate_stoch.py
```diff
@@ -28,30 +28,246 @@
...
+        rise = segment["rise"].to_numpy(dtype=np.float32).reshape(ndip, nstk)
+        tinit = segment["tinit"].to_numpy(dtype=np.float32).reshape(ndip, nstk)
+
+        dstk = float(header["len"]) / nstk
+        ddip = float(header["wid"]) / ndip
+
+        nx = int(np.ceil(float(header["len"]) / dx))
+        ny = int(np.ceil(float(header["wid"]) / dy))
+        wx = _box_average_matrix(nstk, nx, dstk, dx).astype(np.float32)
+        wy = _box_average_matrix(ndip, ny, ddip, dy).astype(np.float32)
```
lispandfound: `wy` uses the same centred padding as `wx`, so the dip-direction overhang is split between the top and bottom of the plane. The stoch header fixes the plane's top edge at `dtop`, with `dhypo` measured from that edge. It records a top-centre point, not a plane centre, so the padded slip ends up shifted down-dip.

**Failure scenario:** Take `wid=3.5`, `dy=2`: `ny=2` and the overhang is 0.25 km at each end. The HF code places row 0 at [dtop, dtop+2] down-dip, but that row's slip was averaged from [-0.25, 1.75] relative to the real top. All slip and rupture times move 0.25 km deeper (up to `dy/2` in general), and `dhypo` no longer lines up with the slip grid. Along strike, centring is correct. Down-dip, all the padding should go at the bottom (offset 0). `test_box_average_matrix_is_centred` and `test_convert_srf_to_stoch_preserves_uniform_slip` lock in the current behaviour.

workflow/scripts/generate_stoch.py
```diff
@@ -60,43 +276,18 @@ def generate_stoch(
...
-        [
-            str(srf2stoch_path),
-            f"dx={dx}",
-            f"dy={dy}",
-            f"infile={srf_ffp}",
-            f"outfile={stoch_ffp}",
-        ]
-    )
+    srf_file = srf.read_srf(srf_ffp)
+    dx = stoch_config.stoch_dx
```
lispandfound: Two special cases were removed: point sources used the SRF grid (`len/nstk`, `wid/ndip`), and other planes were capped at `min(stoch_dx, min_length/2)`. Every plane now uses the 2 km `stoch_dx`/`stoch_dy`, even when the plane is smaller than one cell.

**Failure scenario:** A point-source SRF, or a 0.5 km segment in a multi-fault rupture, becomes a single 2x2 km stoch cell. Moment is conserved, but HF reconstructs a plane up to 4x larger than the real source. Combined with the dip padding, the source centre also sits about `dy/2 * sin(dip)` deeper than it should. If this is intended, it needs a comment. If not, keep a per-plane `dx` for sub-cell planes.

workflow/scripts/generate_stoch.py
```diff
@@ -28,30 +28,246 @@
...
+        rake = segment["rake"].to_numpy(dtype=np.float32).reshape(
+            ndip, nstk
+        ) % np.float32(360.0)
+        rise = segment["rise"].to_numpy(dtype=np.float32).reshape(ndip, nstk)
+        tinit = segment["tinit"].to_numpy(dtype=np.float32).reshape(ndip, nstk)
+
+        dstk = float(header["len"]) / nstk
+        ddip = float(header["wid"]) / ndip
+
+        nx = int(np.ceil(float(header["len"]) / dx))
```
lispandfound: `nx`/`ny` are computed from the header after it is cast to float32 (line 191), so a length that is an exact multiple of `dx` can round up to an extra padded cell.

**Failure scenario:** Checked with numpy: `len=0.3, dx=0.1` gives `ceil(float32(0.3)/0.1) = 4` (float64 gives 3), and `len=1.2, dx=0.4` gives 4 instead of 3. The plane gains a spurious extra cell with a half-cell overhang at each end. With the default `dx=2` and integer lengths it is harmless, but use the float64 header plus a small tolerance, e.g. `ceil(len/dx - 1e-6)`.

workflow/scripts/generate_stoch.py
```diff
@@ -18,7 +18,7 @@
 
 Environment
 -----------
-Can be run in the cybershake container. Can also be run from your own computer using the `generate-stoch` command which is installed after running `pip install ucgmsim-workflow@git+https://github.com/ucgmsim/workflow`. If you are executing on your own computer you also need to specify the `srf2stoch` path (`--srf2stoch-path`).
+Can be run in the cybershake container. Can also be run from your own computer using the `generate-stoch` command which is installed after running `pip install workflow@git+https://github.com/ucgmsim/workflow`.
```
lispandfound: The install instruction was changed to `pip install workflow@git+...`, but `pyproject.toml` names the project `ucgmsim-workflow`.

**Failure scenario:** pip checks the requested name against the package metadata and aborts with an 'inconsistent name' error, so the documented command fails. Keep `ucgmsim-workflow@git+...`.

workflow/scripts/generate_stoch.py
```diff
@@ -28,30 +28,246 @@
...
+
+        stoch_header = StochHeader(
+            longitude=header["elon"],
+            latitude=header["elat"],
+            nx=nx,
+            ny=ny,
+            dx=dx,
+            dy=dy,
+            # The stoch format stores whole degrees.
+            strike=round(header["stk"] % np.float32(360.0)),
```
lispandfound: Strike (and the average rake on line 253) are rounded after the `% 360`, so values in [359.5, 360) are written as `360` instead of `0`.

**Failure scenario:** `stk=359.7` gives `round(np.float32(359.7) % 360) == 360`. Any reader that validates bearings to [0, 360) sees an out-of-range value. Use `round(x) % 360`.

tests/test_generate_stoch.py
```diff
@@ -0,0 +1,408 @@
...
+@pytest.fixture
+def realisation_ffp(tmp_path: Path, synthetic_srf: SrfFile) -> Path:
+    """A realisation whose sources match the planes of the synthetic SRF."""
+    realisation_ffp = tmp_path / "realisation.json"
+    realisations.RealisationMetadata(
+        name="generate stoch test",
+        version="1",
+        defaults_version=defaults.DefaultsVersion.v24_2_2_1,
+    ).write_to_realisation(realisation_ffp)
+    realisations.SourceConfig(
```
lispandfound: The `realisation_ffp` fixture still builds and writes a `SourceConfig` whose planes match the SRF, but `generate_stoch` no longer reads `SourceConfig`.

**Failure scenario:** Dead setup, and the fixture docstring implies a dependency that no longer exists. Drop the `SourceConfig` write.

workflow/scripts/generate_stoch.py
```diff
@@ -28,30 +28,246 @@
...
+
+        dstk = float(header["len"]) / nstk
+        ddip = float(header["wid"]) / ndip
+
+        nx = int(np.ceil(float(header["len"]) / dx))
+        ny = int(np.ceil(float(header["wid"]) / dy))
+        wx = _box_average_matrix(nstk, nx, dstk, dx).astype(np.float32)
+        wy = _box_average_matrix(ndip, ny, ddip, dy).astype(np.float32)
+
+        def box_average(
```
lispandfound: The `box_average` closure is redefined on every loop iteration, with default arguments added only to work around late binding. It wraps a single expression.

**Failure scenario:** Adds indirection for no benefit. Inline `wy @ x @ wx.T`, or move it to a module-level `_box_average(values, wy, wx)`.

workflow/scripts/generate_stoch.py
```diff
@@ -28,30 +28,246 @@
...
+    StochFile
+        An output stoch file downsampled from ``srf_file``.
+    """
+    planes = []
+    for i, segment in enumerate(srf_file.segments):
+        header = srf_file.header.iloc[i].astype(np.float32)
+        nstk, ndip = int(header["nstk"]), int(header["ndip"])
+
+        slip = segment["slip"].to_numpy(dtype=np.float32).reshape(ndip, nstk)
+        rake = segment["rake"].to_numpy(dtype=np.float32).reshape(
```
lispandfound: `rake % 360` is redundant: `circular_mean` works in cos/sin space and already reduces its result with `% 360`.

**Failure scenario:** An extra array operation and more code on every plane. Pass the raw rake values.

lispandfound (commented):
Code review by Claude Code (run locally, since the review action is currently broken). 8 finding(s).

workflow/scripts/generate_stoch.py
lispandfound: The stoch dx behaviour is expected.

PR #129, opened 2026-07-31
Pad nz for free surface boundary condition

lispandfound:
Complement to the velocity modelling PR: https://github.com/ucgmsim/velocity_modelling/pull/59


lispandfound:
Addressed by another merged PR.

PR #130, opened 2026-08-04
Update broadband simulation to use BA18 + CB14 model from site amplification package

lispandfound:
As described, but I took the opportunity to further refactor bb-sim:

1. bb-sim is now dask powered, so requires less memory on clusters by making use of chunking.
2. bb-sim now handles mismatched sample rates between low frequency and high frequency via signal resampling.
3. bb-sim now embeds the vs30 and all the site amplification parameters used into the output.


workflow/default_parameters/root/defaults.yaml
```diff
@@ -1527,3 +1527,9 @@ hf_velocity_model_1d:
       rho: 3.33
       Qp: 394.80
       Qs: 197.40
+bb:
+  fmin: 0.2
+  fmidbot: 0.5
+  fhightop: 100.0
+  fmax: 101.0
+  site_amp_version: "ba2018"
```
lispandfound: This PR removes `fmin`/`fmidbot`/`site_amp_version` from the pinned v24_2_2_1/2/4 (and v26_7_1Hz) defaults and adds a root `bb` block. `defaults.py` merges root into every version, so the older pinned defaults versions silently switch from CB2014 to BA2018 site amplification.

**Failure scenario:** Re-running a realisation with `defaults_version: v24_2_2_2` now gets `site_amp_version: ba2018`, `fmin: 0.2`, `fhightop: 100`, `fmax: 101` from root. Before this PR it got CB2014 with fmin 0.25. Broadband waveforms and IMs change for a defaults version that is meant to reproduce earlier results.

workflow/schemas.py
```diff
@@ -1185,7 +1192,13 @@ def _build_layer(config: dict) -> "LayerConfig":
...
             NUMBER, _is_non_negative
         ),
-        "site_amp_version": str,
+        Literal("fhightop", description="fhightop for site amplification"): And(
+            NUMBER, _is_non_negative
+        ),
+        Literal("fmax", description="fmax for site amplification"): And(
+            NUMBER, _is_non_negative
+        ),
+        "site_amp_version": Use(SiteAmpModel),
```
lispandfound: The BB schema now requires `fhightop`/`fmax` and only accepts `SiteAmpModel` values, so realisations written before this PR (`site_amp_version: "2014"`, no fhightop/fmax) fail with an uncaught `SchemaError`. Nothing falls back or migrates them.

**Failure scenario:** An existing realisation.json has `"bb": {"flo": 1.0, "fmidbot": 0.5, "fmin": 0.25, "site_amp_version": "2014"}`. `BroadbandParameters.read_from_realisation_or_defaults` finds the `bb` key and validates it, which raises SchemaError (fields missing, and "2014" is not a SiteAmpModel). bb-sim crashes on every in-flight or archived realisation.

workflow/default_parameters/root/defaults.yaml
```diff
@@ -1527,3 +1527,9 @@ hf_velocity_model_1d:
       rho: 3.33
       Qp: 394.80
       Qs: 197.40
+bb:
+  fmin: 0.2
+  fmidbot: 0.5
+  fhightop: 100.0
```
lispandfound: The default `fhightop: 100`/`fmax: 101` are at or above Nyquist for typical dt (0.005 s gives Nyquist = 100 Hz), so `amp_highpass` never tapers anything. Full site amplification is applied right up to Nyquist, where the old CB2014 path tapered it off at high frequency.

**Failure scenario:** At bb_dt=0.005 the largest rfft frequency is 100 Hz. The `fft_freqs >= 101` mask is empty and `[100, 101)` only touches the Nyquist bin, so everything above about 15 Hz is now fully amplified. High-frequency IMs (PGA, short-period pSA) shift compared with the previous pipeline, and neither the defaults nor the PR explains why.

workflow/schemas.py
```diff
@@ -1185,7 +1192,13 @@ def _build_layer(config: dict) -> "LayerConfig":
         Literal("fmin", description="fmin for site amplification"): And(
             NUMBER, _is_non_negative
         ),
-        "site_amp_version": str,
+        Literal("fhightop", description="fhightop for site amplification"): And(
```
lispandfound: The `fmin`/`fmidbot`/`fhightop`/`fmax` fields are each validated only as `_is_non_negative`. Nothing enforces `0 < fmin < fmidbot` or `fhightop < fmax`, which `amplification.amp_lowpass`/`amp_highpass` require.

**Failure scenario:** A realisation with `fmin: 0`, or `fmidbot == fmin`, passes validation. bb-sim opens the files, builds the whole dask graph and does I/O, then fails inside the first `_process_bb_chunk` task with a ValueError from `amp_lowpass`, when it could have been rejected at realisation read time.

workflow/scripts/bb_sim.py
```diff
@@ -39,78 +39,285 @@
...
+    # Must drop both waveform variable and time dimension to avoid xarray
+    # automatically reindexing the waveform according to the new axes.
+    new_dset = dset.drop_vars(["waveform", "time"]).assign(waveform=resampled_waveform)
+    new_dset.attrs["dt"] = dt
+    return new_dset
+
+
+# Reference Vs30 (m/s) of the high-frequency simulation, i.e. the Vs30
+# the waveforms are amplified *from* towards each station's target Vs30.
+VS30_SIM = 500.0
```
lispandfound: The reference velocity is hard-coded as `VS30_SIM = 500.0`, but hf-sim already writes the real reference into the HF file as `hf["vref"]` (the 1D model's surface Vs * 1000). The old code also hard-coded 500, but this rewrite is the natural place to use the value that is actually recorded.

**Failure scenario:** A realisation overrides `hf_velocity_model_1d` with a surface layer of Vs 0.8 km/s. hf-sim records vref=800, but bb-sim still amplifies from 500 m/s, so every station's amplification uses the wrong reference and the BB motions come out biased.

workflow/scripts/bb_sim.py
```diff
@@ -39,78 +39,285 @@
...
     """
+    duration = dset["waveform"].sizes["time"] * dset.attrs["dt"]
+    nt = round(duration / dt)
+
+    # NOTE: I am not providing a default start second because we consider it an
+    # error not to provide one (no implicit magic behaviour).
+    new_time = np.arange(nt) * dt + dset.attrs["start_sec"]
+
+    resampled_waveform = xr.apply_ufunc(
+        sp.signal.resample,
```
lispandfound: `resample_signal` uses `scipy.signal.resample`, which is FFT-based and assumes a periodic signal. This PR adds it without any taper or padding, so non-periodic waveforms wrap around and ring (Gibbs).

**Failure scenario:** LF at dt=0.02 is upsampled to the HF dt=0.005. The LF trace ends at a non-zero value, so the jump from its end back to its start leaks ringing into the pre-P-arrival samples and the tail of the broadband record.

workflow/scripts/bb_sim.py
```diff
@@ -39,78 +39,285 @@
...
+        amp = amplification.interpolate_frequencies(amp_model_freqs, fft_freqs, amp)
+        # Constrain the amplification to the [fmin, fmax] band, tapering
+        # logarithmically at either end.
+        amplification.amp_lowpass(fft_freqs, amp, fmin, fmidbot)
+        amplification.amp_highpass(fft_freqs, amp, fhightop, fmax)
+
+        # Taper the tail of the HF waveform (5%) to limit spectral
+        # leakage before amplification.
+        hf_component = hf_waveform[i].copy()
+        amplification.taper(hf_component, 0.05)
```
lispandfound: The 5% tail taper is applied to `hf_waveform[i]` after `align_datasets` has zero-padded it to the common length, not to the HF signal itself.

**Failure scenario:** When the LF record runs past the HF end (for example LF 300 s, HF 200 s), the last 5% of the padded array is already zero. The taper does nothing, and the abrupt HF cut at 200 s goes into the FFT untapered, which is the spectral leakage the comment says it prevents.

workflow/scripts/bb_sim.py
```diff
@@ -145,15 +352,35 @@ def combine_hf_and_lf(
...
+    hf = relabel_hf_components(xr.open_dataset(high_frequency_waveform_file))
+    hf = hf.drop_duplicates("station", keep="first")
+
+    common_stations = sorted(
+        set(map(str, hf.station.values)) & set(map(str, lf.station.values))
+    )
+    # Chunk over stations only, so every chunk holds complete time
+    # series for resampling, alignment and filtering.
+    nt = max(len(lf["time"]), len(hf["time"]))
+    n_stations = round(TARGET_CHUNK_BYTES / (3 * nt * np.float64().itemsize))
```
lispandfound: `n_stations` is computed from `nt` before resampling and alignment, uses `round()` with no `max(1, ...)` guard, and counts only 3 float64 components, while each chunk actually holds LF+HF plus FFT intermediates.

**Failure scenario:** If LF is upsampled (4x for 0.02 to 0.005) or alignment extends the time axis, real chunks are several times `TARGET_CHUNK_BYTES` and workers risk OOM. A very long `nt` makes the value round to 0 and `.chunk` fails. `lf_to_xarray.py` already uses `max(1, TARGET_CHUNK_BYTES // ...)`; use that pattern and compute after resampling.

workflow/scripts/bb_sim.py
```diff
@@ -145,15 +352,35 @@ def combine_hf_and_lf(
...
+    lf = lf.sel(station=common_stations).chunk(chunking)
+    hf = hf.sel(station=common_stations).chunk(chunking)
+
+    bb_dt = min(lf.attrs["dt"], hf.attrs["dt"])
+
+    if not np.isclose(lf.attrs["dt"], bb_dt):
+        lf = resample_signal(lf, bb_dt)
+    if not np.isclose(hf.attrs["dt"], bb_dt):
+        hf = resample_signal(hf, bb_dt)
+    common_stations = sorted(
```
lispandfound: `common_stations` is recomputed and both datasets are `.sel`'d again after resampling. Resampling cannot change the station set and both datasets are already in that order, so lines 383-387 do nothing except add a dask getitem layer after chunking.

**Failure scenario:** No wrong output. It is redundant set building plus extra graph layers after chunking, the pattern the comment at line 355 warns against. Delete lines 383-387.

workflow/scripts/bb_sim.py
```diff
@@ -39,78 +39,285 @@
...
+
+
+LF_COMPONENTS = ["x", "y", "z"]
+"""Component labels every low-frequency output uses.
+
+Written by `lf-to-xarray` for SW4 and by `qcore.timeseries` for EMOD3D:
+x = east-west, y = north-south, z = vertical.
+"""
+
+HF_COMPONENT_TO_LF = {"090": "x", "000": "y", "ver": "z"}
```
lispandfound: `HF_COMPONENT_TO_LF` and the relabel mapping are dead code. This repo's hf-sim already writes `component = ["x", "y", "z"]` (hf_sim.py:344), so the docstring's claim that `hf_simulation` names components 090/000/ver is wrong here.

**Failure scenario:** Maintainers have to keep a mapping and a misleading docstring in sync for a convention no producer in the pipeline emits. The only useful part is the `set(components) != set(LF_COMPONENTS)` check, which can stay on its own.

lispandfound (commented):
Code review by Claude Code (run locally, since the review action is currently broken). 10 finding(s).

